PR #22, opened 2022-07-29
feat: Add flag(s) to filter for file and directory filtering

### LINKED ISSUE #21: [Feature Request] Add flag(s?) to `filter` for file and directory filtering
Currently, passing no choices to `filter` causes it to recursively list every single file, directory, subdirectory, etc. relative to `$PWD`. This is fantastic, however, in parent directories with many (sometimes hundreds or thousands of) files and subdirectories, this can cause `filter` to hang, and also produces output which may very well be useless based on the use-case. 

In light of this, flags related to filtering contents of the current directory might be useful. Some examples:

- `--list-recursive`: Only recurse and show the contents of subdirectories if this flag is passed/true
- `--directories-only`: Only show directories in results

I'll try my hand at a PR implementing the above two flags, as well.

jalavosus:
This commit adds a new flag set to filter.Options, containing (as of
now) one flag: `OnlyDirectories` (`--only-directories`). This flag, when
true, causes `files.List()` to **not** ignore directories, and instead
only return directory names rather than only returning file names.

changes:
- create flag set with options for files.List()
- add files.ListOptions flags to filter.Options

Closes #21

feat: implement ListOptions usage in List

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -8,6 +8,7 @@ import (
 	"github.com/alecthomas/kong"
 	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
+
 	"github.com/charmbracelet/gum/internal/files"
 	"github.com/charmbracelet/gum/internal/stdin"
 	"github.com/charmbracelet/gum/style"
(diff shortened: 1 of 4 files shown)


maaslalani:
Hey! I really appreciate this PR but to avoid adding too many flags for niche use cases I would prefer using `stdin` to handle these cases instead of flags.

```bash
# Non-recursive search
ls | gum filter
# Directories only
ls -d */ | gum filter
```

https://github.com/charmbracelet/gum/issues/21#issuecomment-1200301266

PR #44, opened 2022-07-30
Add soft & hard linter configs & workflows

muesli:
Adds soft- & hard-enforced linter configs and GitHub workflows. Fixed all the linter errors I encountered while I was at it. Please review, in particular the template/renderer error handling. I also noticed there's a bit of unused code, which we may be able to remove (unless it's used by kong's reflection?).


maaslalani:
Thank you @muesli!

PR #62, opened 2022-08-01
Use responsive width

lpmi-13:
Fixed size of terminals was making the mobile view formatting look off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ A tool for glamorous shell scripts. Leverage the power of
 Gloss](https://github.com/charmbracelet/lipgloss) in your scripts and aliases
 without writing any Go code!
 
-<img src="https://stuff.charm.sh/gum/demo.gif" width="100%" alt="Shell running the ./demo.sh script" />
+<img src="https://stuff.charm.sh/gum/demo.gif" width="100%" max-width="600px" alt="Shell running the ./demo.sh script" />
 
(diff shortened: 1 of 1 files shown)


lpmi-13:
Current view on mobile

Screenshot_20220801-094628_Chrome

Mobile view with proposed commit

Screenshot_20220801-094735_Chrome

maaslalani:
Hi! This is great but unfortunately there is no way to specify a `max-width` (at least from what I tried) so the images / GIFs don't look good on desktop. If you know a way to add a max-width and keep the width 100%. Please do let me know!

lpmi-13:
ah right, okay...I feel like I had to deal with that in a previous project. Lemme do some digging and I'll see what I can come up with.

maaslalani:
@lpmi-13 I tried similar things as well, looks like inline-styles don't work due to security reasons:

> The HTML is sanitized, aggressively removing things that could harm you and your kin—such as script tags, inline-styles, and class or id attributes.


https://github.com/github/markup#github-markup

lpmi-13:
Haha, yeah fair enough. I'll go ahead and close this. Fantastic work with this project by the way, gonna see if I can convince my team to start using it!

maaslalani:
Thanks @lpmi-13, really appreciate the PR and you using the tool ❤️

maaslalani:
Please do open issues / let me know if there's anything missing or that you wish was in the tool 🙂

PR #65, opened 2022-08-01
fix: man pages binary description

aymanbagabas:
Sets the right man pages description without any escape sequences and styling.

### DIFF
--- a/man/command.go
+++ b/man/command.go
@@ -12,11 +12,14 @@ import (
 type Man struct{}
 
 // BeforeApply implements Kong BeforeApply hook.
-func (m Man) BeforeApply(app *kong.Kong) error {
-	man := mangokong.NewManPage(1, app.Model)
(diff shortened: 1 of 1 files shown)


man/command.go
```diff
@@ -12,11 +12,14 @@ import (
 type Man struct{}
 
 // BeforeApply implements Kong BeforeApply hook.
-func (m Man) BeforeApply(app *kong.Kong) error {
-	man := mangokong.NewManPage(1, app.Model)
+func (m Man) BeforeApply(ctx *kong.Context) error {
+	ctx.Model.Vars()
+	// Set the correct man pages description without color escape sequences.
+	ctx.Model.Help = "Tasty bubble gum for your shell."
```
maaslalani, line 18:
```suggestion
	ctx.Model.Help = "A tool for glamorous shell scripts."
```

PR #67, opened 2022-08-01
Added option for custom delimiter in filter command

Nickiel12:
I wanted to be able to configure the parsing of the input list for the filter command so it could take a list of space separated values instead of just newline separated values. An example of the new behavior would be:

```SH
echo "candy waffles ice-cream" | ./gum filter --delimiter=" "
> Filter...
• candy   
  waffles 
  icecream
```

### DIFF
--- a/filter/options.go
+++ b/filter/options.go
@@ -13,4 +13,5 @@ type Options struct {
 	Prompt         string       `help:"Prompt to display" default:"> "`
 	PromptStyle    style.Styles `embed:"" prefix:"prompt." set:"defaultForeground=240" set:"name=prompt"`
 	Width          int          `help:"Input width" default:"20"`
+    Delimeter      string       `help:"Delimeter to split input on" default:"\n"`
 }
--- a/filter/options.go
+++ b/filter/options.go
@@ -13,5 +13,5 @@ type Options struct {
 	Prompt         string       `help:"Prompt to display" default:"> "`
 	PromptStyle    style.Styles `embed:"" prefix:"prompt." set:"defaultForeground=240" set:"name=prompt"`
 	Width          int          `help:"Input width" default:"20"`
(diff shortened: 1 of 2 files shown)


filter/options.go
```diff
@@ -13,4 +13,5 @@ type Options struct {
 	Prompt         string       `help:"Prompt to display" default:"> "`
 	PromptStyle    style.Styles `embed:"" prefix:"prompt." set:"defaultForeground=240" set:"name=prompt"`
 	Width          int          `help:"Input width" default:"20"`
+    Delimeter      string       "help:\"Delimeter to split input on\" default:\"\n\""
```
tanyagold, line 16: Did you mean
```suggestion
    Delimeter      string       `help:"Delimeter to split input on" default:"\n"`
```

Nickiel12: If you do that. Than it sets the default delimiter to a literal "\n" instead of a newline

millergarym: Are `"` delimited field tags legal in Go?
Would the following work?
```suggestion
    Delimeter      string       `help:"Delimeter to split input on" default:"
"`
```

Nickiel12: I don't know, but the code I submitted compiled and ran like I expected. Let me try what you put.

Nickiel12: alright, what you posted worked, I didn't want to do that though because now there is a newline that can't be removed without breaking the code - at least the first is explicit, even if it is messy.

justchokingaround:
it would also be nice if `-d` could be used for that

maaslalani:
Btw, in your example you should be able to do:

```
gum filter $(echo candy waffles ice-cream)
```

Since `gum filter` arguments are split by spaces while `stdin` is split by new lines.

PR #89, opened 2022-08-02
Filter --height and Scroll Viewport

### LINKED ISSUE #86: Ability to set maximum height for `gum filter`
I have a script to cd into a repo, but when starting to filter, the list is very long and it will be more aesthetic to limit the search results' height.

A `--height` flag or `--max-height` or something like that.

Non really, using fzf you can do this.

**Additional context**
Navigating with j/k will go off screen if there are too many results

maaslalani:
Fixes #86

### Changes
- Add `--height` option for limiting `gum filter`
- Ability to scroll options (with viewport)

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/alecthomas/kong"
 	"github.com/charmbracelet/bubbles/textinput"
+	"github.com/charmbracelet/bubbles/viewport"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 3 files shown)


maaslalani:
@dlvhdr If you get the chance can you double check that this works correctly for you? It allows `--height x` for `gum filter` and scrolling through options.

dlvhdr:
Works great @maaslalani :)

maaslalani:
Nice! Thank you so much!

PR #90, opened 2022-08-03
feat: Make input and write textarea CharLimit configurable

MathiasGr:
Removes the hardcoded 400 chars CharLimit (default for textarea), and make the CharLimit configurable through an Option.

Fixes #...

### Changes
- Add a `CharLimit` Option to `input` that defaults to the current hardcoded value ( [textarea default CharLimit](https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea.go#L22) )
- Add a `CharLimit` Option to `write` that defaults to the current hardcoded value ( [textarea default CharLimit](https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea.go#L22) )

### DIFF
--- a/input/command.go
+++ b/input/command.go
@@ -28,6 +28,7 @@ func (o Options) Run() error {
 	i.Width = o.Width
 	i.PromptStyle = o.PromptStyle.ToLipgloss()
 	i.CursorStyle = o.CursorStyle.ToLipgloss()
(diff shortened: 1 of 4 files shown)


maaslalani:
Great work, thank you @MathiasGr. I'll get this in soon 🙂

PR #91, opened 2022-08-03
print Style Flags in style command

maplepie:
current style command doesn't print Style Flags in help flag


maaslalani:
Thank you @maplepie!

PR #98, opened 2022-08-04
Satisfy all linters

maaslalani:
### Changes
- Satisfies all linters


maaslalani:
@muesli reverted the `soft` workflow and lint changes as we can have the discussion whether or not to keep / merge those later, but this PR just fixes _all_ of the linting errors / warnings.

PR #106, opened 2022-08-06
Add default timeout on confirm options & messages

### LINKED ISSUE #103: Add auto-confirm timeout
Imagine you need to change some behavior of your script only sometimes, but generally it is fine to run in default configuration. As an example, think about grub bootloader, that boots default configuration unless you change something. 

So, it would be nice for confirm command to have a user-configurable timeout, after which default option would be selected.

ChrHan:
Fixes #103 

### Changes
- Change `confirm` to have a default timeout of 15 seconds
- Configurable timeout using `--timeout` argument 
- Added `help` for options that can be used

This merge request is heavily referenced from [demo of bubbletea](https://github.com/charmbracelet/bubbletea/blob/master/examples/views/main.go), where a timer is implemented to count each second

gum-confirmation-initial

### DIFF
--- a/confirm/command.go
+++ b/confirm/command.go
@@ -17,6 +17,8 @@ func (o Options) Run() error {
 		affirmative:     o.Affirmative,
 		negative:        o.Negative,
 		confirmation:    o.Default,
+		timeout:         o.Timeout,
+		timeoutmsg:      o.TimeoutMsg,
 		prompt:          o.Prompt,
 		selectedStyle:   o.SelectedStyle.ToLipgloss(),
 		unselectedStyle: o.UnselectedStyle.ToLipgloss(),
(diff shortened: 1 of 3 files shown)


TiGR:
What about simplifying interface to

```
   Are you sure?

[ Yes ]  [ No(12) ]
```

ChrHan:
sounds like a good idea, will try to simplify the interface 👍

ChrHan:
gum-confirmation-simplified
here's an updated demo with simplified interface

maaslalani:
Amazing work @ChrHan, I made the `--timeout` flag a `time.Duration` so people can pass any unit of time that they want:

```
gum confirm --timeout 5s
gum confirm --timeout 1m
```

PR #110, opened 2022-08-08
Additional filter subcommand scripting options

dmorgan81:
### Changes

This pull request adds three new options for the filter subcommand to handle common scripting scenarios.

- `--query`: Sometimes a script already has some user input, say an argument to the script itself, that would be useful as an initial filter. `--query`, if supplied, performs an initial filter on the model before the first view.
- `--select-one`: When using a dynamic list or maybe when paired with the aforementioned `--query` option the initial filtered model has just one match. In this case skipping the view and using that one match as the subcommand output can reduce user friction.
- `--exit-zero`: An almost-opposite to `--select-one`. If the initial model, either before or after the initial filter by `--query`, has zero matches treat this as an error condition.

I defaulted `--select-one` and `--exit-zero` to false to preserve backwards compatibility, but I really think they should default to true, at least for `--exit-zero` without an initial filter with `--query`. Filtering an empty list is kinda nonsensical to begin with. I can make this change if the maintainers agree.

There is also room for improvement of the help text. I tried to follow the pithiness of the other options but I had trouble explaining the behavior of `--exit-zero` in a short sentence.

```
$ seq 1 100 | gum filter --query 3
> 3
• 3
  39
  38
  37
  36
  35
  34
  33
  32
  31
  30
  93
  83
  73
  63
  53
  43
  23
  13
```

```
$ seq 1 10 | gum filter --select-one --query 3
3
```

```
$ seq 1 10 | gum filter --exit-zero --query 11
empty list
$ echo $?
1
```

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -58,6 +58,11 @@ func (o Options) Run() error {
 		matches = matchAll(choices)
 	}
 
+	if o.SelectOne && len(matches) == 1 {
+		fmt.Println(matches[0].Str)
+		return nil
+	}
+
 	p := tea.NewProgram(model{
 		choices:        choices,
 		indicator:      o.Indicator,
--- a/filter/options.go
+++ b/filter/options.go
@@ -14,4 +14,5 @@ type Options struct {
 	Width          int          `help:"Input width" default:"20" env:"GUM_FILTER_WIDTH"`
 	Height         int          `help:"Input height" default:"0" env:"GUM_FILTER_HEIGHT"`
 	Value          string       `help:"Initial filter value" default:"" env:"GUM_FILTER_VALUE"`
+	SelectOne      bool         `help:"Auto select single value lists" default:"false" env:"GUM_FILTER_SELECT_ONE"`
 }


maaslalani:
Also, I wonder if we should rename `--query` to `--value` to be consistent with `gum input` and `gum write`, what do you think? I think I could be swayed either way.

dmorgan81:
>Also, I wonder if we should rename --query to --value to be consistent with gum input and gum write, what do you think? I think I could be swayed either way.

I'm all for consistency. I can rename `--query` to `--value`.

> I agree. I don't think we need an `ExitZero` flag either, What do you think about exiting with exit 1 if there is no matches with the query?

That sounds good to be. I also agree with the `grep` analogy.

Should an error message be printed in this case?

maaslalani:
I don't think we need to print an error message, since `grep` does not:

dmorgan81:
Actually I want to expand a bit on the `grep` analogy. I'm not entirely sure if the behavior is correct.

If I have a list of fruits (`echo apple banana grape pear | xargs -n 1 | gum filter`) I can type "orange", find nothing, and revise my query there. In other words it's not an error if the I type something that's not in the list.

Continuing with that flow, if I do type "orange" and hit enter, meaning I've selected nothing, what should the end result be? Currently nothing is printed to stdout and the exit code is zero. Should the `--value` flag mirror that or should both exit with a non-zero exit code? Should the user input be swallowed or should it be passed through even though it wasn't in the list?

maaslalani:
Oh, yeah, that's a _really_ good point. You're right I don't think we should exit if nothing is found, and instead let the user revise the query manually.

maaslalani:
I wonder if we should prevent the user from hitting <kbd>enter</kbd> if there is nothing to select, forcing them to revise the query or press an abort key to exit.

main.go
```diff
@@ -64,6 +64,9 @@ func main() {
 		if errors.Is(err, exit.ErrAborted) {
 			os.Exit(exit.StatusAborted)
 		}
+		if errors.Is(err, exit.ErrEmptyList) {
+			os.Exit(exit.StatusEmptyList)
+		}
```
maaslalani, line 69: I would probably delete this and `exit.StatusEmptyList` and let it fallthrough to the default status of 1.

dmorgan81: The fallthrough would print the error.

maaslalani: Ah totally I forgot! Apologies

maaslalani:
Merged in the `--value` flag as I think that was solid: https://github.com/charmbracelet/gum/pull/113

Still want to discuss the `--select-one` flag since I'm not sure about the functionality / naming 100%

maaslalani:
Also, as a heads up, I made the no options passed as a user error to be consistent with `gum choose`: 
image

dmorgan81:
Just pushed a rebase.

I modeled `--select-one` after the fzf option of the same name. I'm not married to it, though.

maaslalani:
Hey @dmorgan81, going to close this PR since I want to think about it a little bit more before adding this, but I may revisit in the future. Thanks again for the contribution with the default `--value` changes!

PR #112, opened 2022-08-08
feat(choose): add --selected option to choose command

### LINKED ISSUE #72: [feature] Preselected entries in gum choose
Hey everyone!

I just encountered a case where I'd want `gum choose` to display a list of item that **are already selected**. The user is thus prompted to **deselect** some of these items while the default behaviour (just pressing _Enter_) would include all listed items as selected.

This could of course be implemented in the surrounding shell script and by labeling the items something like _Disable XY_ or _Omit XY_, but this results in bad UX, because you're now dealing with a list of negated items.

I have two proposals right now on how to implement this in `gum choose`:

### Escape characters
Introduce some escape character in the item name that is then parsed and leads to this item being preselected.
This allows for fine-grained control, but is error prone due to additional parsing.

### CLI option
Simply add an option like `gum choose --selected` that simply displays all the given items as preselected.

For example, given the command `gum choose --no-limit "foo" "bar" "baz"`, we expect this output:
```
> [•] foo
  [ ] bar
  [ ] baz
```
With an additional CLI option, we could turn this into the following output by calling `gum choose --preselected --no-limit "foo" "bar" "baz"`:
```
> [✕] foo
  [✕] bar
  [✕] baz
```

I'd love to hear some feedback from the maintainers on this. Thanks in advance for your time!

vanielf:
Fixes #72

### Changes
- `--selected` option to indicate what options that should start as selected

### DIFF
--- a/choose/options.go
+++ b/choose/options.go
@@ -13,7 +13,7 @@ type Options struct {
 	CursorPrefix      string       `help:"Prefix to show on the cursor item (hidden if limit is 1)" default:"[•] " env:"GUM_CHOOSE_CURSOR_PREFIX"`
 	SelectedPrefix    string       `help:"Prefix to show on selected items (hidden if limit is 1)" default:"[✕] " env:"GUM_CHOOSE_SELECTED_PREFIX"`
 	UnselectedPrefix  string       `help:"Prefix to show on unselected items (hidden if limit is 1)" default:"[ ] " env:"GUM_CHOOSE_UNSELECTED_PREFIX"`
-	Selected          string       `help:"Options that should start as selected" default:"" env:"GUM_CHOOSE_SELECTED"`
+	Selected          []string     `help:"Options that should start as selected" default:"" env:"GUM_CHOOSE_SELECTED"`
 	CursorStyle       style.Styles `embed:"" prefix:"cursor." set:"defaultForeground=212" envprefix:"GUM_CHOOSE_CURSOR_"`
 	ItemStyle         style.Styles `embed:"" prefix:"item." hidden:"" envprefix:"GUM_CHOOSE_ITEM_"`
 	SelectedItemStyle style.Styles `embed:"" prefix:"selected." set:"defaultForeground=212" envprefix:"GUM_CHOOSE_SELECTED_"`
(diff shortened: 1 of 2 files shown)


willsidlebbc:
Would very much like to see this merged!

vanielf:
I just updated the pull request resolving a conflict that appeared

choose/options.go
```diff
@@ -13,6 +13,7 @@ type Options struct {
 	CursorPrefix      string       `help:"Prefix to show on the cursor item (hidden if limit is 1)" default:"[•] " env:"GUM_CHOOSE_CURSOR_PREFIX"`
 	SelectedPrefix    string       `help:"Prefix to show on selected items (hidden if limit is 1)" default:"[✕] " env:"GUM_CHOOSE_SELECTED_PREFIX"`
 	UnselectedPrefix  string       `help:"Prefix to show on unselected items (hidden if limit is 1)" default:"[ ] " env:"GUM_CHOOSE_UNSELECTED_PREFIX"`
+	Selected          string       `help:"Options that should start as selected" default:"" env:"GUM_CHOOSE_SELECTED"`
```
maaslalani, line 16: I'm fairly sure if you make this a `[]string` `kong` will parse it correctly, allowing you to skip the `parseSelectedItems` step, making the code a little cleaner.

vanielf: you are correct, I just tested it now and it worked! I am sending the changes right now

vanielf:
@maaslalani what do you think about what I said, changing the option name from `-selected` to `--value`, standardizing with the option that was added in `gum filter`?

maaslalani:
@vanielf I think that's quite reasonable, how about we set that as the name and add an alias for `--selected` I think kong has an alias option.

vanielf:
@maaslalani I have no preference between the two options, the idea was just to standardize on the other command
I have little knowledge of go and never used kong before, I looked at the documentation and saw the alias option, added `aliases: "selected"` to the new value option but I couldn't make it work 🙁

maaslalani:
One thing though is that with `gum filter`, `--value` is already being used for the search value I believe so now that there is `gum filter --no-limit` it won't be able to be used to preselect values (so maybe we should stick with `--selected`?).

vanielf:
@maaslalani again no preference, if you think `--selected' is good enough then that's fine with me, you can go ahead 🎉

maaslalani:
Thanks for the great work @vanielf!

PR #113, opened 2022-08-08
Allow --value to pre-fill `gum filter`

maaslalani:
### Changes
- Add `--value` flag to pre-fill `gum filter`


maaslalani:
cc @dmorgan81, wanted to get the `--value` flag in, as I wasn't 100% sure about the naming of `--select-one` yet. But we can continue that discussion.

PR #116, opened 2022-08-14
docs: fix a warning in the filter-key-value example

NalZE7:
### Changes
- Fixing a minor warning ```filter-key-value.sh: line 12: warning: here-document at line 8 delimited by end-of-file (wanted `END')```


maaslalani:
Thank you so much!

PR #123, opened 2022-08-19
added nix flake

IllustratedMan-code:
This is a really simple pull request that adds 2 files `flake.nix` and `flake.lock`. This just makes it easier for people using Nix to get the latest version.

I also added a line to the .gitignore that ignores the default output directory (symlink) when building with nix. 

I would only need to update these files if the dependencies change in the `go.mod` file, but I am happy to maintain it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ testdata
 # Folders
 completions/
 manpages/
+
+# nix
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,25 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1660639432,
+        "narHash": "sha256-2WDiboOCfB0LhvnDVMXOAr8ZLDfm3WdO54CkoDPwN1A=",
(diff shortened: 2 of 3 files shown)


maaslalani:
Just curious if someone wants this why adding an overlay wouldn't be better: https://github.com/maaslalani/_/commit/10ccca0fe049f5175563663a1a621e62485261a3

IllustratedMan-code:
It would just make it easier is all. The user would not have to know how to build go modules or worry about the vendorSha256. 

Once the flake was added to the inputs, gum could be added to the package list like so:

```nix
{inputs, config, pkgs, ... }:
{
environment.systemPackages = [
    inputs.gum.packages.x86_64-linux.default
];
}
```

or if you still wanted an overlay:
```nix
{inputs, config, pkgs, ... }:
let
 gum_overlay = self: super: { gum = inputs.gum.packages.x86_64-linux.default; };
in
{
nixpkgs.overlays = [ gum_overlay ];
}
```

maaslalani:
Sweet, sounds good!

PR #130, opened 2022-08-24
feat(filter): added multiple selections

### LINKED ISSUE #25: [Feature] Allow filter with no limits, like `fzf --multi`.
### Problem

I am trying to achieve something similar to `fzf --multi` functionality where I can pipe/filter output from one program through `gum` for (1) fuzzy-finding and (2) selecting multiple options.

I can accomplish one or the other, like:

- `gum choose --no-limit` but I cannot fuzzy search, or
- `gum filter` but I cannot select multiple options.

### Info

```shell
$ brew info gum
charmbracelet/tap/gum: stable 0.1.0
```

### Repro Steps

```shell
$ echo -e "abc\ndef\nghi\njkl\nmno\npqr\nstu\nvwx\nyz" | gum choose --no-limit
# cannot type anything
```

```shell
$ echo -e "abc\ndef\nghi\njkl\nmno\npqr\nstu\nvwx\nyz" | gum filter
# cannot select multiple options
```

### Possible Solution

Since `gum filter` already has fuzzy finding algorithm implemented, so it might be a relatively easier lift to implement multi-selection with something like `--no-limit` (similar to `gum choose`) and <kbd>TAB</kbd> (not <kbd>SPACE</kbd> since it could be part of the search pattern/string) to select the current line and move the pointer to the next line.

fedeztk:
fix #25

This PR adds the support to multiple selections in `filter` command, more or less like it already happens in `choose`.
When more than one selection is allowed, it is possible to mark/unmark a given number of items by pressing `tab`.
If the number of allowed selections is 1, there is no need to first mark the selection, pressing enter will directly print the selection under the cursor (again, like in `choose`).

 ### Changes
- introduction of flags/Options similar to what is offered by `choose`:
  - `--limit`: set the number of allowed selections (default to 1)
  - `--no-limit`: allows any number of selections, overriding limit
  - `--selected-prefix`: character(s) that mark selected items
  - `--selected-prefix.foreground`: style for the component mentioned above
  - `--unselected-prefix`: character(s) that mark selected items
  - `--unselected-prefix.foreground`: style for the component mentioned above
- introduction of new fields in model struct:
  - `selectedPrefix` and `selectedPrefixStyle`, character(s) and style of selected items, respectively
  - `unselectedPrefix` and `unselectedPrefixStyle`, character(s) and style of unselected items, respectively
  - `limit` and `numSelected` (used with the same purpose found in `choose`)
  - `selected` is the map of currently selected items

I hope the code quality and style is in line with the awesome projects you guys offer. In case anything is wrong I'll happily fix it ASAP.

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -58,16 +58,24 @@ func (o Options) Run() error {
 		matches = matchAll(choices)
 	}
 
+	if o.NoLimit {
+		o.Limit = len(choices)
+	}
+
 	p := tea.NewProgram(model{
-		choices:        choices,
-		indicator:      o.Indicator,
-		matches:        matches,
-		textinput:      i,
-		viewport:       &v,
-		indicatorStyle: o.IndicatorStyle.ToLipgloss(),
-		matchStyle:     o.MatchStyle.ToLipgloss(),
-		textStyle:      o.TextStyle.ToLipgloss(),
-		height:         o.Height,
+		choices:            choices,
+		indicator:          o.Indicator,
+		matches:            matches,
+		textinput:          i,
+		viewport:           &v,
+		indicatorStyle:     o.IndicatorStyle.ToLipgloss(),
+		selectionMarkStyle: o.SelectionMarkStyle.ToLipgloss(),
+		selectionMark:      o.SelectionMark,
+		matchStyle:         o.MatchStyle.ToLipgloss(),
+		textStyle:          o.TextStyle.ToLipgloss(),
+		height:             o.Height,
+		multiSelection:     make(map[string]interface{}),
+		limit:              o.Limit,
 	}, options...)
 
 	tm, err := p.StartReturningModel()
@@ -79,7 +87,14 @@ func (o Options) Run() error {
 	if m.aborted {
 		return exit.ErrAborted
 	}
-	if len(m.matches) > m.selected && m.selected >= 0 {
+
+	// allSelections contains values only if --no-limit is passed,
+	// hence there is no need to check if the option is enabled
+	if len(m.multiSelection) > 0 {
+		for k := range m.multiSelection {
+			fmt.Println(k)
+		}
+	} else if len(m.matches) > m.selected && m.selected >= 0 {
 		fmt.Println(m.matches[m.selected].Str)
 	}
 
--- a/filter/command.go
+++ b/filter/command.go
@@ -88,8 +88,9 @@ func (o Options) Run() error {
 		return exit.ErrAborted
 	}
 
-	// allSelections contains values only if --no-limit is passed,
-	// hence there is no need to check if the option is enabled
+	// allSelections contains values only if limit is greater
+	// than 1 or if flag --no-limit is passed, hence there is
+	// no need to further checks
 	if len(m.multiSelection) > 0 {
 		for k := range m.multiSelection {
 			fmt.Println(k)
--- a/filter/command.go
+++ b/filter/command.go
@@ -63,19 +63,19 @@ func (o Options) Run() error {
 	}
 
 	p := tea.NewProgram(model{
-		choices:            choices,
-		indicator:          o.Indicator,
-		matches:            matches,
-		textinput:          i,
-		viewport:           &v,
-		indicatorStyle:     o.IndicatorStyle.ToLipgloss(),
-		selectionMarkStyle: o.SelectionMarkStyle.ToLipgloss(),
-		selectionMark:      o.SelectionMark,
-		matchStyle:         o.MatchStyle.ToLipgloss(),
-		textStyle:          o.TextStyle.ToLipgloss(),
-		height:             o.Height,
-		multiSelection:     make(map[string]interface{}),
-		limit:              o.Limit,
+		choices:                choices,
+		indicator:              o.Indicator,
+		matches:                matches,
+		textinput:              i,
+		viewport:               &v,
+		indicatorStyle:         o.IndicatorStyle.ToLipgloss(),
+		selectedIndicatorStyle: o.SelectedIndicatorStyle.ToLipgloss(),
+		selectedIndicator:      o.SelectedIndicator,
+		matchStyle:             o.MatchStyle.ToLipgloss(),
+		textStyle:              o.TextStyle.ToLipgloss(),
+		height:                 o.Height,
+		multiSelection:         make(map[string]interface{}),
+		limit:                  o.Limit,
 	}, options...)
 
 	tm, err := p.StartReturningModel()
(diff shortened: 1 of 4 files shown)


maaslalani:
Thank you so much for this PR! This is fantastic ❤️

filter/options.go
```diff
@@ -4,14 +4,18 @@ import "github.com/charmbracelet/gum/style"
...
-	PromptStyle    style.Styles `embed:"" prefix:"prompt." set:"defaultForeground=240" envprefix:"GUM_FILTER_PROMPT_"`
-	Width          int          `help:"Input width" default:"20" env:"GUM_FILTER_WIDTH"`
-	Height         int          `help:"Input height" default:"0" env:"GUM_FILTER_HEIGHT"`
-	Value          string       `help:"Initial filter value" default:"" env:"GUM_FILTER_VALUE"`
+	Indicator          string       `help:"Character for selection" default:"•" env:"GUM_FILTER_INDICATOR"`
+	IndicatorStyle     style.Styles `embed:"" prefix:"indicator." set:"defaultForeground=212" envprefix:"GUM_FILTER_INDICATOR_"`
+	Limit              int          `help:"Maximum number of options to pick" default:"1" group:"Selection"`
+	NoLimit            bool         `help:"Pick unlimited number of options (ignores limit)" group:"Selection"`
+	SelectionMark      string       `help:"Character for selection marks" default:">" env:"GUM_FILTER_SELECTED"`
+	SelectionMarkStyle style.Styles `embed:"" prefix:"selection-mark." set:"defaultForeground=212" envprefix:"GUM_FILTER_SELECTED_"`
```
maaslalani, line 12: Can we change `SelectionMark` to `SelectedIndicator`?

```suggestion
	SelectedIndicator      string       `help:"Character to indicate selected items" default:">" env:"GUM_FILTER_SELECTED_INDICATOR"`
	SelectedIndicatorStyle style.Styles `embed:"" prefix:"selected-indicator." set:"defaultForeground=212" envprefix:"GUM_FILTER_SELECTED_INDICATOR_"`
```

filter/command.go
```diff
@@ -58,16 +58,24 @@ func (o Options) Run() error {
...
+		matches:            matches,
+		textinput:          i,
+		viewport:           &v,
+		indicatorStyle:     o.IndicatorStyle.ToLipgloss(),
+		selectionMarkStyle: o.SelectionMarkStyle.ToLipgloss(),
+		selectionMark:      o.SelectionMark,
+		matchStyle:         o.MatchStyle.ToLipgloss(),
+		textStyle:          o.TextStyle.ToLipgloss(),
+		height:             o.Height,
+		multiSelection:     make(map[string]interface{}),
```
maaslalani, line 77: I think we should change this to `map[string]bool`, what do you think?

```suggestion
		multiSelection:     make(map[string]bool),
```

filter/filter.go
```diff
@@ -22,18 +22,23 @@ import (
...
-	aborted        bool
-	quitting       bool
-	matchStyle     lipgloss.Style
-	textStyle      lipgloss.Style
-	indicatorStyle lipgloss.Style
+	textinput          textinput.Model
+	viewport           *viewport.Model
+	choices            []string
+	matches            []fuzzy.Match
+	selected           int
```
maaslalani, line 29: We should probably rename `selected` to `cursor` since I guess that better represents what state it is holding.

fedeztk: I agree with that. I was a bit unsure about naming variables in general.

filter/options.go
fedeztk: Seems reasonable to me, I'll change it.

filter/command.go
fedeztk: I don't see the benefits on using a boolean since it's basically meaningless which values the map holds. Anyways I don't have any strong opinion about it and I might miss the purpose behind this, thus I can make a new commit with that change reflected on the code.

maaslalani: Hmm, I see your point. Do you know if there are memory benefits of having this `map` map to pointless values rather than booleans? I would guess that there may be some which might be slightly advantageous with large filters.

fedeztk: To my surprise it actually is counter-productive to use the interface{}. On second thought, interface{} can effectively hold any value, and therefore here is the explanation on its slowness.

I'll leave here the result of my simple benchmark.
```
goos: linux
goarch: amd64
pkg: bench
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkInterface-16            4549672               251.9 ns/op
BenchmarkEmptyStruct-16          6298020               172.5 ns/op
BenchmarkBool-16                 5262289               190.8 ns/op
PASS
ok      bench   4.955s
```
Test file:
```go
package main

import "testing"

func BenchmarkInterface(b *testing.B) {
	m := map[int]interface{}{}
	for i := 0; i < b.N; i++ {
		m[i] = nil
	}
}

func BenchmarkEmptyStruct(b *testing.B) {
	m := map[int]struct{}{}
	for i := 0; i < b.N; i++ {
		m[i] = struct{}{}
	}
}

func BenchmarkBool(b *testing.B) {
	m := map[int]bool{}
	for i := 0; i < b.N; i++ {
		m[i] = true
	}
}
```

maaslalani: Sweet! Looks like we should be using an empty `struct{}` then? I think that makes sense to me. Thank you for running these benchmarks!

maaslalani: Worth also just double checking if theres a difference between setting the boolean to `false` rather than `true` in case that makes any difference.

fedeztk: Good point on the false vs true possible difference.
I just tested it and it did not seem to make any difference:
```
goos: linux
goarch: amd64
pkg: bench
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkInterface-16            3932340               293.8 ns/op
BenchmarkEmptyStruct-16          6849700               166.6 ns/op
BenchmarkBool-16                 6222806               176.5 ns/op
PASS
ok      bench   6.053s
```

maaslalani: Awesome! Thank you so much, I say we should use an empty `struct{}` in that case

fedeztk: Perfect, I'll work on a commit that fixes all the changes proposed in the review then.

maaslalani: ❤️

maaslalani (commented):
Hey @fedeztk, this PR is so fantastic. Just playing around with some styling and I think I was able to come up with this. What do you think about it?

image

maaslalani (approved):

fedeztk:
This looks stunning! Thanks for taking the time to improve and review this PR.

PR #134, opened 2022-08-28
docs: add repology badge and add install command for Termux

pvonmoradi:
- `gum` is packaged for Termux.
- Added repology badge
- Split debian add-source command into two lines

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,9 @@ gum confirm "Commit changes?" && git commit -m "$SUMMARY" -m "$DESCRIPTION"
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


muesli:
Just my two cents: I'm not a big fan of the repology badge, as it becomes a bit unwieldy very quickly. As discussed previously, we could move the installation instructions into a separate document. The badge probably fits better there, too.

maaslalani:
Thanks for the contribution @pvonmoradi, going to close this for now while we think about how to better manage installation instructions / badges!

PR #139, opened 2022-08-31
make git-branch-manager.sh example executable

MichalGrzymek:
Since this script has a shebang line and also all other shell examples are executable, it should also be executable


maaslalani:
Thank you @MichalGrzymek! Much appreciated!

PR #140, opened 2022-09-01
feat(choose): add optional flag

### LINKED ISSUE #87: Add --optional flag to `choose` command
Not totally sure if this is a bug or if this should be a feature request.  When using the `choose --no-limit` command it always sends through whichever row is highlighted regardless of whether it has been selected.  

Add a `--optional` flag and only return options that are selected (as denoted by an X in the box with default style) when this flag is used.  If no options are selected then return an empty string.

The alternative is to bake a blank option into the choices and then add logic to look for result to match this option after `gum choose` operation.

**Additional context**
```
cat ~/choice_example.txt | gum choose --no-limit
> [•] one
  [ ] two
  [ ] three
one
```
Example of it returning whatever the line lands on without explicit selection.

fedeztk:
Fixes #87 

### Changes
- added `--optional` flag so that without explicitly options selected, `enter` doesn't select highlighted entry

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -74,6 +74,7 @@ func (o Options) Run() error {
 		cursorPrefix:      o.CursorPrefix,
 		items:             items,
 		limit:             o.Limit,
+		optional:          o.Optional,
 		paginator:         pager,
 		cursorStyle:       o.CursorStyle.ToLipgloss(),
 		itemStyle:         o.ItemStyle.ToLipgloss(),
@@ -98,7 +99,7 @@ func (o Options) Run() error {
 		}
 	}
 
-	fmt.Println(strings.TrimSuffix(s.String(), "\n"))
+	fmt.Print(s.String())
 
 	return nil
 }
--- a/choose/choose.go
+++ b/choose/choose.go
@@ -29,6 +29,7 @@ type model struct {
 	quitting         bool
 	index            int
 	limit            int
+	optional         bool
 	numSelected      int
 	paginator        paginator.Model
 	aborted          bool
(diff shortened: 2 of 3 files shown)


choose/command.go
```diff
@@ -98,7 +99,7 @@ func (o Options) Run() error {
 		}
 	}
 
-	fmt.Println(strings.TrimSuffix(s.String(), "\n"))
+	fmt.Print(s.String())
```
fedeztk, line 102: I want to point out that, whether this PR will be accepted or not, this is still very reasonable (unless I'm missing something). Trimming the last "\n" and than adding it back again by using `fmt.Println` instead of just using `fmt.Print` does nothing but calls one additional function. 

In the case of this specific PR anyways, I added this because otherwise, when the `--optional` flag was passed, an extra new line was printed when nothing was selected, resulting in a wrong return ("\n" instead of "").

maaslalani: That's a very good point! Thank you, we will definitely change this.

maaslalani:
This PR is awesome! Thank you so much @fedeztk. Let me think about a bit whether we want to add this flag or if we simply want to remove the functionality of having at least one selection altogether in multiselects (i.e. always `--optional`)

choose/command.go
maaslalani: This change is now on `main`: https://github.com/charmbracelet/gum/commit/81602545a65795d094fe7873d3548c22576f75b6

fedeztk: Awesome! Thanks for the feedback, I'm glad this was added.

MikaelFangel:
@maaslalani - Any news of whether this feature will be merged or the old functionality will be removed?

maaslalani:
@MikaelFangel Thanks for the reminder on this one.

@fedeztk I've made it so that the default multiple select is always optional.

maaslalani:
https://github.com/charmbracelet/gum/commit/e6de7749b139c3ff38285437d390f5f92e1689fb

fragolinux:
> @MikaelFangel Thanks for the reminder on this one.
> 
> @fedeztk I've made it so that the default multiple select is always optional.

hi, if this is merged, it does not seem to work...
```
gum --version
gum version v0.9.0 (832c4fc)
gum choose --no-limit 1 2 3 # i just press enter, no selection
1
```

MikaelFangel:
@fragolinux be aware 0.9.0 is not the latest release. 😇

image

fragolinux:
good! just tried, all works as expected now, thanks!

PR #147, opened 2022-09-09
fix(format): do not append additional new line

### LINKED ISSUE #146: Gum format --type emoji adds new line
**Describe the bug**
When using `gum format --type emoji` new line is added to original string

**To Reproduce**
1. `gum format --type emoji :+1:  | wc -l`
4. `2` is shown as the output

Another example:
```bash
$ gum format --type emoji This is a string without emojis && echo second line
This is a string without emojis

second line
$ 
```

**Expected behavior**
`gum format --type emoji :+1:  | wc -l`  should return 1 as there is only one line in the input string

Desired output from previous example:
```bash
$ gum format --type emoji This is a string without emojis && echo second line
This is a string without emojis
second line
$ 
```
**Screenshots**

 - OS: Ubuntu 22.04
 - Gum: v0.6.0

fedeztk:
Fixes #146 

### Changes
- replaced `fmt.Println` with `fmt.Print`, as discussed in #146

### DIFF
--- a/format/command.go
+++ b/format/command.go
@@ -41,6 +41,6 @@ func (o Options) Run() error {
 		return err
 	}
 
-	fmt.Println(v)
+	fmt.Print(v)
 	return nil
 }


maaslalani:
Thank you so much!

PR #160, opened 2022-09-27
Updates hash and adds overlay to flake

solomon-b:
Fixes #...

The hash was out of date in the nix flake. I went to update it and went ahead and made some other basic improvements in the flake.

1. [flake-utils](https://github.com/numtide/flake-utils) is a commonly used nix library which makes it easier to build flakes which target all the default systems supported by Hydra. 

2. Adding an Overlay allows nix users to more quickly merge `gum` into their local nixpkgs. 

### Changes
- Updates the vendor hash in the Nix Flake
- Uses flake-utils to support more systems in Nix
- Adds a NIx overlay.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,7 @@
+{ pkgs }:
+
+pkgs.buildGoModule {
+  name = "gum";
+  src = ./.;
+  vendorSha256="rMqhYZMa0+5F3X4WDm4jE6IwlzOugqm65SAP38bdQx8=";
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,21 +1,39 @@
 {
   "nodes": {
-    "nixpkgs": {
+    "flake-utils": {
       "locked": {
-        "lastModified": 1660639432,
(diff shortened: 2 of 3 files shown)


maaslalani:
Thank you so much! I really appreciate this PR!

solomon-b:
@maaslalani would you like nix flake support added to other charmbracelet projects?

maaslalani:
@solomon-b I would hold off for now, we want to see if it is worth supporting nix flakes in other projects by using `gum` as a bit of a test to see how much extra maintenance and benefit it is!

solomon-b:
I'm not actually a `go` developer so I don't know what tooling you would use in your development environment but this flake should include a development shell as well. Maybe someone with more `go` experience can add that later (or tell me what you want included and i can add it).

IllustratedMan-code:
@solomon-b Thanks for doing this! I was the original creator of the flake and I was considerably less experienced with nix than I am now when I made it. I really appreciate the improvements. 

The flake does appear to be broken again, think it just needs a vendorsha256 update, but I was wondering what you thought about using https://github.com/nix-community/gomod2nix to make updating the hash a little easier.

PR #164, opened 2022-10-02
feat: add location to spin

gregkh:
Sometimes it is nice to spin to the right side of the title, not the left, so add a --location option to pick the left or right side.  The default remains "left".

New option is:
	--location="left|right"
and the environment variable GUM_SPIN_LOCATION can also be used.

### DIFF
--- a/spin/options.go
+++ b/spin/options.go
@@ -11,4 +11,5 @@ type Options struct {
 	SpinnerStyle style.Styles `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
 	Title        string       `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
 	TitleStyle   style.Styles `embed:"" prefix:"title." envprefix:"GUM_SPIN_TITLE_"`
+	Align        string       `help:"Alignment of spinner with regard to the title" short:"a" type:"align" enum:"left,right" default:"left" env:"GUM_SPIN_ALIGN"`
 }
--- a/spin/spin.go
+++ b/spin/spin.go
@@ -25,6 +25,7 @@ import (
 type model struct {
 	spinner spinner.Model
 	title   string
+	align   string
 	command []string
 	aborted bool
 
@@ -73,7 +74,13 @@ func (m model) Init() tea.Cmd {
 		commandStart(m.command),
 	)
 }
-func (m model) View() string { return m.spinner.View() + " " + m.title }
+func (m model) View() string {
+	if m.align == "left" {
+		return m.spinner.View() + " " + m.title
+	}
+
+	return m.title + " " + m.spinner.View()
+}
 
 func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	var cmd tea.Cmd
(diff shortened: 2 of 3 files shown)


gregkh:
naive attempt to move the spinner location to be at the end of the line.  If you all have a better suggestion for the command line argument, or wish to use the --style or --title options somehow (I didn't want to overload them), that's fine with me, just a proposed way to do this, thanks!

spin/options.go
```diff
@@ -11,4 +11,5 @@ type Options struct {
 	SpinnerStyle style.Styles `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
 	Title        string       `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
 	TitleStyle   style.Styles `embed:"" prefix:"title." envprefix:"GUM_SPIN_TITLE_"`
+	Location     string       `helo:"Location of spinner" short:"l" type "location" enum:"left,right" default:"left" env:"GUM_SPIN_LOCATION"`
```
muesli, line 14: Small typo: `helo` -> `help`.

spin/spin.go
```diff
@@ -73,7 +74,13 @@ func (m model) Init() tea.Cmd {
 		commandStart(m.command),
 	)
 }
-func (m model) View() string { return m.spinner.View() + " " + m.title }
+func (m model) View() string {
+	if (m.location == "left") {
```
muesli, line 78: This might seem odd coming from C, but in Go you tend to avoid the parentheses if there's only a single conditional. I'll push a fix to your fork.

spin/options.go
muesli: Also noticed a missing colon after the `type` struct tag. Fixing.

spin/spin.go
muesli: Pro tip: in the Go world there's a standard tool called `gofmt` that automatically formats your sources according to the language conventions. It even fixes little Go characteristics like the missing parentheses.

More about it here: https://go.dev/blog/gofmt

I realize you're just dabbling in our Go code to post this quick fix, but I figured it can't harm to point this out :smile:

muesli:
I guess an alternative could be naming it `Align`, which other `gum` commands already offer with similar options.

spin/options.go
gregkh: argh, sorry about that, and thanks for fixing.

spin/spin.go
gregkh: > This might seem odd coming from C, but in Go you tend to avoid the parentheses if there's only a single conditional. I'll push a fix to your fork.

I much prefer this style (it's the kernel style to not have a { } for only a single conditional), I wasn't sure about the Go style, sorry.

gregkh:
> I guess an alternative could be naming it `Align`, which other `gum` commands already offer with similar options.

That's fine with me as well, it's your call.  I can redo it that way if you want (with your fixes as well.)  Just let me know.

muesli:
Yeah, let's go with `Align` for consistency's sake. Feel free to squash the commits down locally or I can do it during the merge.

gregkh:
Ok, will squash down and change to align and resubmit here tomorrow, thanks so much for the quick review!

gregkh:
That was simpler, now pushed a new version.

maaslalani:
Thank you so much for the contribution @gregkh! We really appreciate it!

PR #166, opened 2022-10-02
feat(choose): Preselect the Selected Option When Limit is One

mathaou:
Fixes #...

### Changes
- 
- 
-

Added starting index to preselect the selected option when limit is one


maaslalani:
Thanks so much @mathaou, I pushed a change to make the `hasSelectedItems` work with `limit == 1` as well so that we don't have to re-query the `array.Contains` call. This looks great and will merge soon!

PR #169, opened 2022-10-03
feat(file): support symlinks

aymanbagabas:
Highlight symlinks and display their resolved path.


maaslalani:
Thank you so much! You're the best! ❤️

PR #172, opened 2022-10-05
Add file path option to table

mathaou:
Fixes #...

### Changes
- 
- 
-


maaslalani:
I might be missing something but how come this is a `--flag`:

Why is file reading:
```
gum table --file example.csv
```

better than reading from stdin:
```
gum table < example.csv
```

mathaou:
> I might be missing something but how come this is a `--flag`:
> 
> Why is file reading:
> 
> ```
> gum table --file example.csv
> ```
> 
> better than reading from stdin:
> 
> ```
> gum table < example.csv
> ```

It's moreso about fidelity. I was hooked up to a debugger and testing with a 10k line .csv file, so it was easier just to add a file option than it was to launch the binary + stdin redirection and then attach the debugger each time. I was thinking about other CLI tools like awk/grep that allow for both as well.

maaslalani:
Fair point, I can agree with that. I will merge this PR.

PR #175, opened 2022-10-06
fix(confirm): an empty --negative will not be displayed

### LINKED ISSUE #145: OK Dialog
In some cases we want to inform the user before performing some task. 
For example when a task is going to take a really long time, we might want to tell the user beforehand, so they don't get impatient and interrupt the scripts prematurely. 

Adding a simple "OK" dialog with some text would be helpful in these cases. 
This could be its own command or an alteration of  the `confirm` command (e.g. with a `--type="OK"` or `--show-negative=0` ).

Currently we need to either use `gum choose "OK"` or `gum confirm --affirmative="OK" --negative=""`, both of which are suboptimal solutions.

Here is an example of the dialog we would like to model:

keevan:
This change will hide the normally "No" button when the text is empty. In most cases, this will serve as an acknowledgement and the affirmative message would likely change.

This seems to be better than displaying an empty button.

Resolves #145

### DIFF
--- a/confirm/confirm.go
+++ b/confirm/confirm.go
@@ -103,6 +103,11 @@ func (m model) View() string {
 		neg = m.selectedStyle.Render(m.negative + timeout)
 	}
 
+	// If the option is intentionally empty, do not show it.
+	if m.negative == "" {
+		neg = ""
+	}
+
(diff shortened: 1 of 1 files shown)


maaslalani:
Nice work @keevan, this is a great solution for #145!

PR #176, opened 2022-10-06
feat(write): use --header to display a label above the textarea

### LINKED ISSUE #80: Introduce a way to have descriptive text before or after `gum write` UI
I would like a way to provide header and/or footer text around the `gum write` UI. The other input commands let you provide a prompt of some sort. But with `gum write`, the prompt is repeated for each line, and isn't useful as informative text.

My specific use case is that it isn't obvious that you have to send an EOF (ctrl-D) to indicate when you're done editing the text. So I wanted to have some text either before or after the editing area (maybe using `--faint` and `--italic`) to tell people to ctrl-d when done. Other uses could be having some descriptive text that tells them what they're editing.

I could just `print` it out before calling `gum write` but then it'll hang around in the terminal history which is ugly. I like how the other prompts like for `gum input` and `gum choose` go away.

My current solution is to abuse `gum choose` with an empty choice as an "interstitial" they have to pass through before being shown `gum write`:

```
gum choose --cursor.italic --cursor.faint --cursor="
Press return to edit your script, and ctrl-d to save" ""
```

Not ideal, as it introduces another step.

Ideally I'd be able to do `gum write --intro "Press ctrl-d to save" --intro.faint --intro.italic`

(Really don't care what it's called... `intro` and `outro` could work. `header` and `footer` could work.

keevan:
Resolves #80

### DIFF
--- a/write/command.go
+++ b/write/command.go
@@ -47,7 +47,11 @@ func (o Options) Run() error {
 	a.SetHeight(o.Height)
 	a.SetValue(o.Value)
 
-	p := tea.NewProgram(model{textarea: a}, tea.WithOutput(os.Stderr))
+	p := tea.NewProgram(model{
+		textarea:    a,
+		header:      o.Header,
+		headerStyle: o.HeaderStyle.ToLipgloss(),
+	}, tea.WithOutput(os.Stderr))
 	tm, err := p.StartReturningModel()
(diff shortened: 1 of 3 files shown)


maaslalani:
Really nice work @keevan! Thanks so much for this great PR.

Issue #114, opened 2022-08-09
gum filter --reverse


caarlos0:
**Is your feature request related to a problem? Please describe.**
fzf's default style is prompt in the bottom, list in the top... gum's is the other way around.,

**Describe the solution you'd like**
fzf also has a `--layout reverse` flag to make it the other way around... I believe we need something like that in for `gum filter` too.

**Describe alternatives you've considered**
haven't found any alternatives

**Additional context**

maaslalani:
TIL about `fzf --layout reverse`! Yeah, we can have something similar in Gum :)

dhruvmanila:
`fzf` offers multiple layouts:
```
*        --layout=LAYOUT
*               Choose the layout (default: default)
  
                default       Display from the bottom of the screen
                reverse       Display from the top of the screen
                reverse-list  Display from the top of the screen, prompt at the bottom
```

The default for `gum` is the `reverse` layout. By `reverse`, @caarlos0 do you mean the `default` layout?

caarlos0:
> The default for `gum` is the `reverse` layout. By `reverse`, @caarlos0 do you mean the `default` layout?

yes, so, my suggestion is to add the "alternate" layout, do we can have a `--reverse` flag as well... but in gum's case, gum's default == fzf's `--reverse`, and gum's `--reverse` == fzf's default.l

PR #177, opened 2022-10-06
feat(filter): add reverse layout

dhruvmanila:
resolves: #114 

### Changes

Add `--reverse` flag to `gum filter` command to display the list in reverse direction similar to the default layout for `fzf`

filter_reverse

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -78,6 +78,7 @@ func (o Options) Run() error {
 		height:                o.Height,
 		selected:              make(map[string]struct{}),
 		limit:                 o.Limit,
+		reverse:               o.Reverse,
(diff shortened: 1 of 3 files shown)


maaslalani:
Hey, this looks almost good to go. One last case we need to handle is when you are in `--reverse` and scroll up past the viewport and then filter (by typing more text) the cursor will be off the screen. We would need to clamp it to be inside the viewport.

To reproduce:

```
gum filter --reverse --height 10
```

Press Up 20 times

Type: `Dock`

The cursor will be off the screen

maaslalani:
Awesome work @dhruvmanila, thank you so much for the contribution!

PR #180, opened 2022-10-08
perf: speed up filter rendering

muesli:
We don't need to style each character individually, buffer them and flush before highlights. Greatly improves performance when styling thousands of items after each cursor movement.


maaslalani:
Thanks so much @muesli! This is great ❤️

PR #181, opened 2022-10-08
fix(choose): error if selected options > limit

dhruvmanila:
**Rationale:** If the number of selected options in `--selected` flag is greater than the limit, the last `n` options are selected where `n` is the limit.

Fixes #...

### Changes

Throw an error if the number of selected options are greater than the specified limit


maaslalani:
Thank you @dhruvmanila

PR #186, opened 2022-10-08
feat(filter): added `--exact` for exact matches

### LINKED ISSUE #50: Option to filter by whole words (not fuzzy)
For instance

echo "The quick brown fox" | gum filter

Will show up if I type `hioo` in the filter.

I'd prefer an option where it will only show up `quick` or `k bro`

enrichman:
Fixes #50 

### Changes
Added the `--exact` flag to perform an exact filter (not fuzzy).

To highlight the correct letters I had to override the ones matched originally, and also I had to sort the slice again to avoid the "shuffling" of some elements that were ordered with the fuzzy ranking algorithm.

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -11,7 +11,6 @@
 package filter
 
 import (
-	"sort"
 	"strings"
 
 	"github.com/charmbracelet/bubbles/textinput"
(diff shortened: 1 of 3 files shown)


Justinzobel:
Ooh nice work thanks @enrichman!

filter/filter.go
```diff
@@ -180,6 +182,11 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			// https://github.com/sahilm/fuzzy
 			m.matches = fuzzy.Find(m.textinput.Value(), m.choices)
 
+			// For exact search return only the exact matches
+			if m.exact {
+				m.matches = exactMatches(m.textinput.Value(), m.matches)
+			}
```
dhruvmanila, line 188: I think we can put the fuzzy finding in the else block here as it's either exact or fuzzy.

enrichman: Ah, yes, right! We can filter directly the `choices`. And since they are stable I've also removed the sort. 👍

maaslalani:
@enrichman would you be okay if the flag was `--fuzzy` and `--no-fuzzy`

maaslalani:
This is how it looks, if someone uses both flags than it defaults to `--no-fuzzy`.

image

enrichman:
Ah, that's pretty neat! Thanks for sharing.

maaslalani:
Fantastic work on this PR, this was a heavily requested feature! Thank you so much. Will merge after CI is green 😄

PR #194, opened 2022-10-12
updated sha256 in default.nix

IllustratedMan-code:
Dependencies in the `go.mod` file were updated which broke the nix derivation. This fixes it.
### Changes
- Update sha256 in `default.nix` to match new dependencies.


maaslalani:
Thank you!

PR #222, opened 2022-10-24
docs: picture tag is causing display issues

muesli:
Often fails displaying on Firefox & mobile Chrome.


maaslalani:
Thanks @muesli, works great!

PR #229, opened 2022-11-04
fix(table): output proper CSV for the selected row

lmarlow:
### Changes
- `gum table` outputs proper CSV

#### Columns with commas in them

Before:
table-pr-before-1

After:

table-pr-after-1

#### Columns with escaped quotes

Before:
table-pr-before-2

After:
table-pr-after-2

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -4,7 +4,6 @@ import (
 	"encoding/csv"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/alecthomas/kong"
 	"github.com/charmbracelet/bubbles/table"
@@ -102,7 +101,11 @@ func (o Options) Run() error {
 	}
 
 	m := tm.(model)
-	fmt.Println(strings.Join([]string(m.selected), o.Separator))
+
+	w := csv.NewWriter(os.Stdout)
+	w.Comma = reader.Comma
+	w.Write([]string(m.selected))
+	w.Flush()
 
 	return nil
 }
--- a/table/command.go
+++ b/table/command.go
@@ -105,8 +105,7 @@ func (o Options) Run() error {
 
 	m := tm.(model)
 
-	err = writer.Write([]string(m.selected))
-	if err != nil {
+	if err = writer.Write([]string(m.selected)); err != nil {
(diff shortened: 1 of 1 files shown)


table/command.go
```diff
@@ -102,7 +101,11 @@ func (o Options) Run() error {
 	}
 
 	m := tm.(model)
-	fmt.Println(strings.Join([]string(m.selected), o.Separator))
+
+	w := csv.NewWriter(os.Stdout)
+	w.Comma = reader.Comma
+	w.Write([]string(m.selected))
```
muesli, line 107: We can probably just `return w.Write(...)` here. This also resolves the linter error.

table/command.go
```diff
@@ -102,7 +101,11 @@ func (o Options) Run() error {
 	}
 
 	m := tm.(model)
-	fmt.Println(strings.Join([]string(m.selected), o.Separator))
+
+	w := csv.NewWriter(os.Stdout)
+	w.Comma = reader.Comma
```
muesli, line 106: Even though `reader.Comma` will contain the right value, I would probably still set this to `o.Separator`. The code above might change in the future and suddenly `reader.Comma` isn't initialized correctly.

muesli (changes requested):

table/command.go
lmarlow: f6d23cd

table/command.go
lmarlow: `Flush` needs to be called or it won't actually print.

table/command.go
```diff
@@ -102,7 +104,13 @@ func (o Options) Run() error {
 	}
 
 	m := tm.(model)
-	fmt.Println(strings.Join([]string(m.selected), o.Separator))
+
+	err = writer.Write([]string(m.selected))
+	if err != nil {
```
muesli, line 109: Just one final nitpicky request: collapse the conditional into a single line.

lmarlow: c951467 does it, I think. Go isn't my main language. Thanks for the pointers.

muesli (approved):
LGTM!

maaslalani (approved):
Thank you!

PR #232, opened 2022-11-09
feature(choose): Allow user to input additional value like in ComboBox

deicon:
Enhancement #...
Adding Option to convert List into ComboBox allowing the user to enter a custom value

### Changes
- --allow-input Option enabling Input Field 

image

image

image


deicon:
@maaslalani A similar feature as in #233. Did you have a chance to have a alook at them as well ?

Issue #213, opened 2022-10-19
gum tag support for gum filter


keevan:
**Is your feature request related to a problem? Please describe.**
I'm looking to add the functionality of tag like support for e.g. gum filter. This would allow me to select an option already available OR accept whatever input I currently have entered.

**Describe the solution you'd like**
When I type in the input, it filters down like it would normally using `gum filter`, but if there is no match. OR I decide that the current input is what I'm actually after (despite various alternatives listed below), then I can choose to apply the selected input instead.

This behaviour would more or less be similar to the one available for things such as adding labels to an github issue. 

Another cool thing that might be added later on (ooh scope creep) is to auto suggest the next word, based on connecting words in the current listing.

For example:
Given three options:
- I would like to sit down on the beach.
- I would like to go to sleep
- I am thinking going to the movies.

Then:
- When I type something like: "I would like", it filters to the top 2 options.
- Bonus points: When I type "Now I'm going to", it would almost predict the next word I'm thinking about in the input field (if that's possible), and I might be able to cycle through the options, such as "sit", "go" or "the". In terms of real world usage, this could be very useful for note taking and related tasks and workflows. I can imagine you might want to enter similar but different things. Though I can see one could argue that this could instead option in a text editor (similar to adding a git commit via vim) where autocomplete might be natively supported in the editor. Yeah still juggling this particular thought.

**Describe alternatives you've considered**
I was thinking about having `gum tag` as its own command. It might be confusing having them as one type of command once more tag related features are added, but there is quite a big overlap in features I believe, so opting for least path of resistance for now.

**Additional context**
N/A

maaslalani:
We were just discussing adding something like this today! This would be an awesome feature.

I think we should add this to `gum filter` (since 95% of the functionality would be shared). What do you think about:

```
gum filter --custom
gum filter --allow-custom
```

keevan:
Yes, get the basics added first :)

Agreed, it should be available on gum filter. 

Of the two, I would lean towards the more verbose --allow-custom, as the other one is a bit more vague. 

Thanks.

brunocbreis:
What if it's just a matter of outputting whatever the user typed, even if there's not a match?
I'm thinking something like a default value of `--strict=true`, meaning it would output nothing if there is no match, and `--strict=false`, which toggles this behavior of outputting a custom option.
If `--strict=false` and `--fuzzy` is enabled, it would prioritize a fuzzy find if there is one. Otherwise, it returns whatever it got from the user.

[...]

PR #233, opened 2022-11-09
feature(choose): Adding option to use non-matching filter as result

deicon:
Small enhancement to the filter to allow a non matching filter to be a valid 
result. 

### Changes
New Option --no-strict to take filter string if no match was found

image

image

image

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -87,7 +87,6 @@ func (o Options) Run() error {
 		return fmt.Errorf("unable to run filter: %w", err)
 	}
 	m := tm.(model)
-
 	if m.aborted {
 		return exit.ErrAborted
 	}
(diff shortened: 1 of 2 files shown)


maaslalani:
Hey, thank you so much for this PR, it is exactly what we wanted in #213. I think we'll need to consider other names for the flag though. Perhaps `--allow-any`, `--allow-input`, `--any-input`, `--freeform`, `--unrestricted` etc...

maaslalani:
Maybe, `--constrained` and `--unconstrained` or `--constraints` and `--no-constraints`

maaslalani:
Also maybe `--strict` and `--strict=false` / `--no-strict` would be great like @brunobe mentioned

deicon:
I agree with the property names and I would prefer 

--allow-input 
or
--strict

maaslalani:
> I agree with the property names and I would prefer
> 
> --allow-input or --strict

Let's go with `--strict` 👍

deicon:
Updated to --strict (default) and --no-strict allowing additional input

keevan:
Thanks so much for adding this @deicon :smiley:

mainrs:
I just opened #239, which I think this PR covers. Would it be possible to get a new tagged release so package maintainers can update properly?

PR #148, opened 2022-09-10
fix(prompt): fix default selection of prompt, timer and exit code

ChrHan:
Fixes #142 

### Changes
- Change timeout to show on correct prompt (not only on `exit code 1`)
- Fix timeout to not enforce user options to `exit code 1`

Demo for 3 commands:
- [x] `❯ ./gum confirm --timeout 3s --default=true "delete"` -> should return `exit code 0` and default to `yes`
- [x] `❯ ./gum confirm --timeout 3s --default=false "delete"` -> should return `exit code 1` and default to `no` 
- [x] `❯ ./gum confirm --timeout 3s "delete"` -> should return `exit code 0` and default to `yes`
gum-confirm-timeout-fix

Issue #142, opened 2022-09-05
gum confirm --default="yes" --timeout=3s 'delete' 


ziwen1002:
**Describe the bug**

run command
 `gum confirm --default="yes" --timeout=3s 'delete'`
- The timer is always on the label of NO
- The default action after the overtime is NO

**To Reproduce**
1. run command  `gum confirm --default="yes" --timeout=3s 'delete'`

**Expected behavior**
- The timer is on the label of default
- The default action after the overtime is default

**Screenshots**
none

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04
 - Browser none
 - Version gum version v0.5.0 (0834690)

**Smartphone (please complete the following information):**
none

**Additional context**
none

ChrHan:
If I understand correctly, you would like to set default to "yes" when timeout ends.

In that case, command should be `gum confirm --timeout=3s 'delete'` as the `--default` parameter only accept `true` for `exit code 0` or `false` for `exit code 1`

However, looking at the codebase, there can be an improvement made to not enforce `gum confirm --timeout` to `exit code 1` - I'll post the proposed PR later on

ziwen1002:
Yes, this is what I expected.

ziwen1002:
> If I understand correctly, you would like to set default to "yes" when timeout ends.
> 
> In that case, command should be `gum confirm --timeout=3s 'delete'` as the `--default` parameter only accept `true` for `exit code 0` or `false` for `exit code 1`
> 
> However, looking at the codebase, there can be an improvement made to not enforce `gum confirm --timeout` to `exit code 1` - I'll post the proposed PR later on

Yes, this is what I expected

muesli:
It would also be nice to have a `--timeout` flag for commands like `choose`, `input`, and `spin`.

PR #260, opened 2023-01-08
featiure: Adding timeout parameter to file, choose, input, filter, pa…

### LINKED ISSUE #218: Support --timeout and --default flag for more commands
Commands like choose, input, and spin should also support the `--timeout` and `--default` flags.

deicon:
Adding Timeout flag to automatically abort running command

Fixes #218 #142 
Incorporates #148 

### Changes
- Use provided default in Confirm when timeout happens
- refactor timeout option into common 
- Refactor Timeout Ticker 
- Make TimeoutMsg to contain current timeout value.


deicon:
@maaslalani Anything missing I should add to this ?

maaslalani:
@deicon I will give this a review today! Thanks again for this awesome PR.

timeout/options.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+// CmdOptions Common Timeout Option.
+type CmdOptions struct {
```
maaslalani, line 11: This can probably be named `Options` or `TimeoutOptions` unless I am missing something!

choose/options.go
```diff
@@ -1,10 +1,13 @@
 package choose
 
-import "github.com/charmbracelet/gum/style"
+import (
+	"github.com/charmbracelet/gum/style"
+	common "github.com/charmbracelet/gum/timeout"
```
maaslalani, line 5: We can probably leave this as `timeout` import I believe. `timeout.Options` is nice as the option unless there is a reason I am missing as to why it is named as such.

choose/command.go
```diff
@@ -106,6 +106,8 @@ func (o Options) Run() error {
 		itemStyle:         o.ItemStyle.ToLipgloss(),
 		selectedItemStyle: o.SelectedItemStyle.ToLipgloss(),
 		numSelected:       currentSelected,
+		hasTimeout:        o.HasTimeout(),
+		timeout:           o.Timeout,
```
maaslalani, line 110: I believe for `gum choose` we should either make it select:
1. The first option always
2. The option the user's cursor is on when the timeout hits
3. Exit with status 1 to indicate timeout

Currently, the `gum confirm 1 2 3 --timeout 2s` will timeout with status 0 (success) but not have any output which could be misleading, so we should either exit with status 1 or return some output.

input/input.go
```diff
@@ -19,24 +22,42 @@ type model struct {
...
+		timeStr = timeout.Str(m.timeout)
+	}
 	if m.header != "" {
 		header := m.headerStyle.Render(m.header)
-		return lipgloss.JoinVertical(lipgloss.Left, header, m.textinput.View())
+		return lipgloss.JoinVertical(lipgloss.Left, header, m.textinput.View()+" "+timeStr)
 	}
 
-	return m.textinput.View()
+	return timeStr + " " + m.textinput.View()
```
maaslalani, line 48: IMO we should remove this time indicator as it is not consistent and not customizable, we can add a customizable timer later!

maaslalani (commented):
Great work! Some comments that need updating but this should be good to merge soon!

timeout/options.go
deicon: a struct in a package timeout ist not supposed to be named anything starting with timeout. 
It is supposed to read timeout.CmdOptions. Instead of repeating timeout.TimeoutOptions. 

So Options would be fine. But,.. there was one place in the choose command Options, where this would clash as the Name Options was already used in the struct.
By the way, this struct also does not read to easily. 
A struct called Options containing a field Options. 

````
package choose

import (
	"github.com/charmbracelet/gum/style"
	common "github.com/charmbracelet/gum/timeout"
)

// Options is the customization options for the choose command.
type Options struct {
     Options []string `arg:"" optional:"" help:"CmdOptions to choose from."`

    common.CmdOptions              // including timeout command options [Timeout,...]
````

choose/options.go
deicon: yes. common is wrong. Agreed. But as said above, timeout.Options clashes with the choose commands Options struct

choose/command.go
deicon: I think you made a typo in the command. It is supposed to be choose, right ? If so, I would agree on selecting the first/selected option when timeout hits

input/input.go
deicon: Removing it for input only or always ? I would like to see them for confirm for example. But you are right, ist not styleable nor customizable.

choose/command.go
maaslalani: Yes typo! I meant gum choose!

timeout/options.go
deicon: IMHO The more golangey way would be to use this as a means of inheriting common probs w/o having the indirection through an extra property. But sure, for readability this is also an option.

deicon: Renaming Options in the choose Options struct is not doable, as this would break existing user scripts

deicon: just renamed and pushed

choose/options.go
deicon: changed

input/input.go
maaslalani: `gum confirm` is okay since it's a bit of a special case but just remove it for `gum input` imo since it seems a bit out of place

choose/options.go
maaslalani: Thank you!

timeout/options.go
maaslalani: Awesome! I agree, reads much better. Thank you so much for changing!

deicon:
Adjusted to your comments.

deicon:
Anything left here ?

deicon:
So, is there still need for this ? If not I cancel the PR. If its still wanted, i will update conflicts. But only if needed as I said. Otherwise it seems like time waste.

rpodgorny:
@maaslalani could you please take a look at this? i am unable to switch my current scripts to gum without this feature. thanks!

deicon:
> @maaslalani could you please take a look at this? i am unable to switch my current scripts to gum without this feature. thanks!

I just fixed conflicts. But as far as I am concerned for the last time :)

deicon:
Resolved merge conflicts

deicon:
> @maaslalani could you please take a look at this? i am unable to switch my current scripts to gum without this feature. thanks!

I have resolved the conflicts again. Lets see

maaslalani:
Thanks so much @deicon, I also just resolved merge conflicts again after some recent merges. Would you be open to introducing this feature with a PR for each subcommand. This PR is a bit hard to review and I think it would be much easier to test if we could introduce the timeout flag to each subcommand individually.

I think one PR to introduce the `timeout.go` stuff and then a PR for each subcommand would be ideal. I totally understand if you don't want to break this down into multiple PRs, If that's the case, I can do that myself :)

Thanks again for this PR, it's awesome!

deicon:
@maaslalani I will have a look at it end of this week. If possible, I can split it up. Will let you know latest Friday.

maaslalani:
Thanks! No rush on it, I really appreciate all your work on this one.

deicon:
I am splitting this up into simple PRs per command

deicon:
@maaslalani I hope I did not mess something up by splitting up commits. Please check and review the PRs.

maaslalani:
@deicon Thanks so much for splitting up the PRs, made it much easier to review, test, and merge them in. I really appreciate all the effort ❤️. Thank you again!

PR #264, opened 2023-01-14
An easier way to do this right?

olivertzeng:
Fixes #...

### Changes
- 
- 
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ Available spinner types include: `line`, `dot`, `minidot`, `jump`, `pulse`, `poi
 Select a row from some tabular data.
 
 ```bash
-gum table < flavors.csv | cut -d ',' -f 1
+gum table -s , < flavors.csv
 ```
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey! This is slightly different in terms of what functionality the first command tries to achieve. The first one allows you to choose any row of the table and then takes the first column of it to use in the next pipeline where as the second command will simply output the rows.

First output:

```
Orange
```

Second output:

```
Orange,$0.25
```

maaslalani:
Thank you for this PR, we really appreciate it! Thanks for looking for potential improvements in the README! ❤️

olivertzeng:
> Hey! This is slightly different in terms of what functionality the first command tries to achieve. The first one allows you to choose any row of the table and then takes the first column of it to use in the next pipeline where as the second command will simply output the rows.
> 
> First output:
> 
> ```
> Orange
> ```
> 
> Second output:
> 
> ```
> Orange,$0.25
> ```

whoa thats really useful, thanks!

PR #269, opened 2023-01-28
Add staged file check for commit example

GeekMasher:
I found out that I had to [add these 3 lines to my `commit` script](https://github.com/GeekMasher/.dotfiles/blob/main/dev/.local/commit) using Gum to make sure if I didn't commit anything I would be prompted if I wanted to include all of the unstaged files.

The big issue with this snippet is that it uses `grep` and `awk` so if a user doesn't have these of their system, it would fail.

I don't know if there is a better way to do this so I'm leaving it like this.

### Changes

- Added check to commit example in the case of no staged commits


maaslalani:
Can we move this check to the first line of the script?

1. Check if there are any unstaged files.
2. Ask to "Stage Files?"

PR #270, opened 2023-01-30
feat(file): Add page up/down key bindings

Stoux:
Add support for going up/down a 'page' in `gum file`.

[asciicast](https://asciinema.org/a/w7MT9F2w2IoFziusn7VxofUya)


maaslalani:
Great work @Stoux, thank you so much!

PR #271, opened 2023-02-02
not choose something when hitting enter with `gum choose` 

olivertzeng:
I don't know how to program with go so please help modify it ok? I want a flag called `--allow-unchosen`


MikaelFangel:
This duplicates #140 🙂

PR #293, opened 2023-03-07
fix: for using --value in combination w/ --no-fuzzy

### LINKED ISSUE #292: `--value` causes `--no-fuzzy` to be ignored
**Describe the bug**
When the `--value` option is given, the `--no-fuzzy` option is ignored.

**To Reproduce**
```
$ echo "~/src/tomcat\n~/Nextcloud/Action" | gum filter --no-fuzzy --value="cat"
```

**Expected behavior**
When the `--value` and `--no-fuzzy` options are given, they both should be honored.

 - OS: macOS 13.2.1 on Apple Silicon
 - gum version: 0.9.0

MikaelFangel:
Fixes #292

### Changes
- Added a switch to check for the combination of the two flags
- Kept the original behaviour of using fuzzy if the --no-fuzzy flag isn't present

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -53,8 +53,13 @@ func (o Options) Run() error {
 	var matches []fuzzy.Match
 	if o.Value != "" {
 		i.SetValue(o.Value)
+	}
+	switch {
+	case o.Value != "" && o.Fuzzy:
 		matches = fuzzy.Find(o.Value, choices)
-	} else {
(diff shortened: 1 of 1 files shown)


maaslalani (approved):
This is perfect, thanks so much for catching this one.

MikaelFangel:
You're welcome just glad I can contribute to the project 🎉

PR #300, opened 2023-03-09
fix: table ignores foreground color set by the user

### LINKED ISSUE #294: gum table ignores color overrides for selected.foreground
**Describe the bug**
setting $GUM_TABLE_SELECTED_FOREGROUND and /or--selected.foreground="212" is always ignored and is set to 212

**To Reproduce**

echo 'Bubble Gum,Price,Ingredients
Strawberry,$0.88,"Water,Sugar"
Guava,$1.00,"Guava Flavoring,Food Coloring,Xanthan Gum"
Orange,$0.99,"Sugar,Dextrose,Glucose"
Cinnamon,$0.50,"Cin""na""mon"'|gum table  --selected.foreground="13" 

**Expected behavior**
color is set to whatever is designated in parameter

**Screenshots**
image

 - OS: windows / rocky linux
 - version: 0.9.0

MikaelFangel:
Fixes #294

### Changes
- Removed the inheritance function because when setting the selected style it would not allow changes to already set values. The default colour remains unchanged, and now it's the user's choice whether the selected text should be bold or not.

This can also be fixed by removing the setter of the foreground colour on this [line](https://github.com/charmbracelet/bubbles/blob/1ca90b3c942b5833c53a6fdfef78cd1fddfe6025/table/table.go#L100) in the bubbles project.

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -74,7 +74,7 @@ func (o Options) Run() error {
 	styles := table.Styles{
 		Cell:     defaultStyles.Cell.Inherit(o.CellStyle.ToLipgloss()),
 		Header:   defaultStyles.Header.Inherit(o.HeaderStyle.ToLipgloss()),
-		Selected: defaultStyles.Selected.Inherit(o.SelectedStyle.ToLipgloss()),
+		Selected: o.SelectedStyle.ToLipgloss(),
 	}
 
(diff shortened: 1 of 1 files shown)


maaslalani (approved):
Thank you so much, this looks great!

PR #303, opened 2023-03-10
feat(Spin): Option to show live output

hopefulTex:
### Changes Spin
- Added option to view live output when using spin command
- Added `--live-output` flag

### DIFF
--- a/spin/options.go
+++ b/spin/options.go
@@ -6,7 +6,8 @@ import "github.com/charmbracelet/gum/style"
 type Options struct {
 	Command []string `arg:"" help:"Command to run"`
 
-	ShowOutput   bool         `help:"Show output of command" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
+	ShowOutput   bool         `help:"Show output of command after its completion" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
+	LiveOutput   bool         `help:"Show output of command in realtime" default:"false" env:"GUM_SPIN_LIVE_OUTPUT"`
 	Spinner      string       `help:"Spinner type" short:"s" type:"spinner" enum:"line,dot,minidot,jump,pulse,points,globe,moon,monkey,meter,hamburger" default:"dot" env:"GUM_SPIN_SPINNER"`
 	SpinnerStyle style.Styles `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
 	Title        string       `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
--- a/spin/options.go
+++ b/spin/options.go
@@ -6,8 +6,7 @@ import "github.com/charmbracelet/gum/style"
 type Options struct {
 	Command []string `arg:"" help:"Command to run"`
 
-	ShowOutput   bool         `help:"Show output of command after its completion" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
-	LiveOutput   bool         `help:"Show output of command in realtime" default:"false" env:"GUM_SPIN_LIVE_OUTPUT"`
+	ShowOutput   bool         `help:"Show output of command during execution" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
 	Spinner      string       `help:"Spinner type" short:"s" type:"spinner" enum:"line,dot,minidot,jump,pulse,points,globe,moon,monkey,meter,hamburger" default:"dot" env:"GUM_SPIN_SPINNER"`
 	SpinnerStyle style.Styles `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
 	Title        string       `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
--- a/README.md
+++ b/README.md
@@ -294,8 +294,7 @@ gum pager < README.md
 Display a spinner while running a script or command. The spinner will
 automatically stop after the given command exits.
 
-To view the command's output after completion, use the `--show-output` flag.
-To view output in real-time use the `--live-output` flag.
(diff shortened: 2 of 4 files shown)


hopefulTex:
I think that the `--show-output` flag should effect both live and accumulated outputs, but it's also not very intuitive to need two flags to do one thing and one flag to do a similar thing.

maaslalani:
I think it makes sense for this to simply replace the `--show-output` behaviour. Likely people passing `--show-output` and pipeing to a file to capture the output so it shouldn't make a difference in that case and otherwise IMO it looks better coming in like.

hopefulTex:
@bashbunni bump

spin/options.go
```diff
@@ -6,7 +6,8 @@ import "github.com/charmbracelet/gum/style"
 type Options struct {
 	Command []string `arg:"" help:"Command to run"`
 
-	ShowOutput   bool         `help:"Show output of command" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
+	ShowOutput   bool         `help:"Show output of command after its completion" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
```
bashbunni, line 9: Looks like Maas said you can outright replace ShowOutput with LiveOutput, should be able to merge once that's done!

bashbunni (changes requested):
Replace ShowOutput with new LiveOutput

spin/options.go
hopefulTex: <--The Point--
      Me
Will do!

hopefulTex:
Okay, just to be clear,
The desired behavior when using the `--show-output` flag is:
Display the commands' output AS IT IS GENERATED.
Display the `Stderr` AFTER EXECUTION.
If this *is* the case, then would it be worth ripping out `Stdout` from the `finishCommandMsg` or is it best to leave it in for possible future changes?

maaslalani:
@hopefulTex I think the `--show-output` flag can simply return the commands output and stderr as it is generated. No need to buffer the text.

hopefulTex:
I struggled on whether you meant the literal keywords `buffer` and `return` or not.
I hope this is what you meant ;w;

`--show-output` flag now spits out `stdout` and `stderr` of the running program as new text is added to the buffer

maaslalani:
That sounds good, basically we should just spit out text live anytime we receive it rather than hold on to it.

maaslalani:
@hopefulTex, I believe we'll need to separate the spinner from the output. I.e. the output should not necessarily be part of the Bubble Tea view since I could want to show the spinner while my output pipes into another file.

```
gum spin --show-output ./some-script > log.txt
```

In this case, we'll want to show the spinner and output the text / output to `os.Stdout` (which is a reason why the spinner runs on `stderr`).

hopefulTex:
Alright, hopefully this one does the trick.
I'm about to implement:
- spinner -> `stderr`
- output -> `stdout`
- only display `stdout` if `--show-output`
Hopefully correct :?

hopefulTex:
I was so caught up in the fact that I could I never stopped to think if I should.
I suppose we can grant the people a little POSIX compliance, as a treat.

maaslalani:
Sorry, to clarify we should still send the output to stdout since this output could be useful to the next program in the pipeline, but since the user is piping it they likely don't want to see it.

hopefulTex:
@maaslalani The requested change has gone through, though I am conflicted about the output handling with pipes.
When piping to a file, with no `--show-output` flag, the resulting file is empty because nothing is spit out.
I feel like it's not very intuitive to need a "show" output flag to silently pipe the output to a file, but at the same time it offers greater control without requiring a new flag.
On the other- other hand, maintaining the same behavior of previous versions is probably best for all the gum pioneers.

hopefulTex:
Closed until this bug is figured out:
`gum spin -- gum confirm` adds a new line every tea.Msg

hopefulTex:
There are so many issues that I feel as though interactivity is a little out of scope for this pr at the least.
@maaslalani am I good to pretend interactive programs don't exist?

bdrnglm:
Does that mean that this feature is never going to be implemented ?

hopefulTex:
> Does that mean that this feature is never going to be implemented ?

I'm sorry, it will be difficult for me to find the time to polish this until the semester comes to a close, this should be out by the end of May at the latest!

Mikrz:
> @maaslalani The requested change has gone through, though I am conflicted about the output handling with pipes. When piping to a file, with no `--show-output` flag, the resulting file is empty because nothing is spit out. I feel like it's not very intuitive to need a "show" output flag to silently pipe the output to a file, but at the same time it offers greater control without requiring a new flag. On the other- other hand, maintaining the same behavior of previous versions is probably best for all the gum pioneers.
> 
> EDIT:: Just refreshed for your latest comment, dw I understood 👍 (I hope?)

It might be worth checking out how https://github.com/sharkdp/bat handles piping, as it's output method is detected on the fly by default. If you run just `bat <file>` it prints the formatted output, but if it's piped it automatically (by default) prints the raw file just like normal cat does. It is written in Rust so naturally the code is going to be different, but perhaps it can help provide a guide in the right direction?

I feel like the default behavior of either dumping process output live or buffered in a pipeline would make sense. Given this functionality, I think a --[no-]buffer-pipe flag of some sort would make sense, to change the behavior from `tail -f` style output to the current --show-output behavior of dumping output at the end.

hopefulTex:
Given the area of effect for compatibility with interactive programs, and my tendency for scope-creep, this pr should just cover simple live output.
My apologies for the long wait.

hopefulTex:
@maaslalani Using the `--show-output` flag now displays program output in realtime and outputs through `stdout` when piping.
I believe this feature is (finally) complete. :)

bashbunni (approved):
Tested locally both writing to TTY and piping to a file. I used the following to test:
```bash
gum spin --show-output --spinner minidot --title "Checking for outdated brew packages" -- brew outdated > outdated.txt
```
and it looked great. Will still get @maaslalani to double check everything looks the way he wants before merging :)

PR #306, opened 2023-03-13
removed bksp from stdin

### LINKED ISSUE #297: Pager doesn't render man pages properly
`gum pager` seems to not render the `|` on the right side properly.

Using gum 0.9.0:
```
export PAGER="gum pager"
man tar
```

Screenshot from 2023-03-08 15-33-05

MikaelFangel:
Fixes #297

### Changes
- Removes backspaces, including the character before it, using a regex. The problem stems from when setting the variable and calling the pager, you get the man pages in their nroff format. Which means that it includes ^H chars, and I suspect that hinders bubbletea from calculating the width correctly.

### DIFF
--- a/pager/command.go
+++ b/pager/command.go
@@ -2,12 +2,12 @@ package pager
 
 import (
 	"fmt"
-
 	"github.com/alecthomas/kong"
 	"github.com/charmbracelet/bubbles/viewport"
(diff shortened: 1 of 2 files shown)


maaslalani:
Thank you so much, this works great!

MikaelFangel:
You're welcome! (:

PR #316, opened 2023-03-20
fix: added esc to keys and changed error code to 130

### LINKED ISSUE #315: `gum file` should terminate with exit code 130 if the user presses escape
**Describe the bug**
In gum 0.9.0 `gum filter` and `gum choose` both terminate with exit code 130 if the user presses escape. `gum file` does not accept escape as input. As far as I can tell the only way to exit `gum file` without choosing a file is to press Control-C.

**To Reproduce**
```
$ gum file /
```
Then try to press escape. Nothing happens

**Expected behavior**
When you press escape in `gum file`, it should terminate with exit code 130.

MikaelFangel:
Fixes #315

### Changes
- Make escape also quit `file`
- Change the returning error code to 130

I tried to make the behaviour as similar as the one from `filter`/`chose`.

### DIFF
--- a/file/file.go
+++ b/file/file.go
@@ -20,6 +20,7 @@ import (
 type model struct {
 	filepicker   filepicker.Model
 	selectedPath string
+	aborted      bool
 	quitting     bool
 }
 
@@ -31,7 +32,9 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch msg.String() {
-		case "ctrl+c", "q":
+		case "ctrl+c", "q", "esc":
+			m.aborted = true
+			m.quitting = true
 			return m, tea.Quit
 		}
 	}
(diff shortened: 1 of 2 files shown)


file/file.go
```diff
@@ -31,7 +32,9 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch msg.String() {
-		case "ctrl+c", "q":
+		case "ctrl+c", "q", "esc":
```
maaslalani, line 35: I believe `esc` is used to go back when a user navigated into a directory.

```suggestion
		case "ctrl+c", "q":
```

MikaelFangel: Isn't that a bit confusing when most gum programs cancel on escape? And you already can go back using the arrow keys.

maaslalani: That's a fair point! I think that makes sense to me. I like the consistency and because there is already a way to go up / back. We can merge this as-is.

PR #317, opened 2023-03-20
fix: changed logic in stdin and stopped ignoring err in filter

### LINKED ISSUE #313: `gum filter` hangs on certain input pipes closing without sending any options
**Describe the bug**
In gum version 0.9.0, `gum filter` hangs if certain input pipes close without sending any options

**To Reproduce**
From the shell type (here ^D means to type Control-D):
```
$ cat - | gum filter
line one
line two
^D
```
As expected, gum shows two lines and lets you choose between them.

Now try:
```
cat - | gum filter
^D
```
`gum` just hangs waiting for more input, you have to `^C` to kill it.

**Expected behavior**
If you do:
```
cat - | gum filter
^D
```
`gum` should do what `echo "" | gum filter` does
```
echo | gum filter
no options provided, see `gum filter --help`
```
and exit with exit code 1

MikaelFangel:
Fixes #313

### Changes
- Change logic in stdin to look for EOF or err
- Check when existing for loop if string is empty
- Check error in the `filter` command

Before:
demo

After:
demo

### DIFF
--- a/internal/stdin/stdin.go
+++ b/internal/stdin/stdin.go
@@ -28,10 +28,6 @@ func Read() (string, error) {
 		}
 	}
 
-	if len(b.String()) <= 0 {
-		return "", fmt.Errorf("stdin is empty")
-	}
-
 	return b.String(), nil
 }
 
--- a/internal/stdin/stdin.go
+++ b/internal/stdin/stdin.go
@@ -19,13 +19,17 @@ func Read() (string, error) {
 
 	for {
 		r, _, err := reader.ReadRune()
-		if err != nil && err == io.EOF {
+		if err != nil || err == io.EOF {
 			break
 		}
 		_, err = b.WriteRune(r)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


maaslalani:
I think if the following is run, it should not say `stdin is empty`, it should find files in the current directory to not have a breaking change in behaviour:

```
gum filter
```

MikaelFangel:
I agree, in that it shouldn't be a breaking change, and have reverted all the breaking changes. So now it only changes the logic listening for EOF.

internal/stdin/stdin.go
```diff
@@ -19,7 +19,7 @@ func Read() (string, error) {
 
 	for {
 		r, _, err := reader.ReadRune()
-		if err != nil && err == io.EOF {
+		if err != nil || err == io.EOF {
```
maaslalani, line 22: Correct me if I'm wrong but we probably should `continue` rather than `break` if `err != nil` and `break` if `err == io.EOF`?

PR #318, opened 2023-03-20
fix: correct wrong key combination for gum write in README.md

### LINKED ISSUE #312: gum write
**Describe the bug**
The gum ver：0.9.0（832c4fc）
# The manual is :
Write:
Prompt for some multi-line text.
Note: CTRL+D and esc are used to complete text entry. CTRL+C will cancel.
```gum write > story.txt```

when use esc the result is the same with ctrl+c ,can't save the input mes,ctrl + d is useful

MikaelFangel:
Fixes #312

### Changes:
 - Moves the `esc` over to the cancel option together with CTRL+c, so it reflects the current behaviour of the program.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ gum input --password > password.txt
 
 Prompt for some multi-line text.
 
-Note: `CTRL+D` and `esc` are used to complete text entry. `CTRL+C` will cancel.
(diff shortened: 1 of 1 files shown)


maaslalani:
Thank you so much!

PR #321, opened 2023-03-22
feature: add pager search functionality

### LINKED ISSUE #308: Implement search functionality for gum pager
When I normally use pagers like `less` I almost use the search functionality every time, but I can't do this with gum as it is now.

The ability to search when using `gum pager`. Maybe by pressing `/` as with `less` and `vim`.

MikaelFangel:
Fixes #308 

### Changes
- [x] Highlight all matches in the viewport
- [x] Add input field for regex pattern
- [x] Create new truncate function for lipgloss
- [x] Support going to next/prev using `n` and `p`
- [x] Create key handler
- [x] Update help text
- [x] Go to previous match using N
- [x] Replace help text with search when active
- [x] Create function to calculate Lipgloss padding from render
- [x] Support the search with soft-wrap enabled
- [x] Indicate what the current match is
- [x] Self review and clean-up code

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -1,6 +1,10 @@
 package utils
 
-import "github.com/charmbracelet/lipgloss"
+import (
+	"strings"
+
+	"github.com/charmbracelet/lipgloss"
+)
 
 // LipglossTruncate truncates a given line based on its lipgloss width.
 func LipglossTruncate(s string, width int) string {
@@ -8,6 +12,14 @@ func LipglossTruncate(s string, width int) string {
 	for i = 0; i < len(s) && lipgloss.Width(s[:i]) < width; i++ {
 	} //revive:disable-line:empty-block
 	return s[:i]
+}
+
+// LipglossLengthPadding calculated calculates how much padding a string is given by a style.
+func LipglossLengthPadding(s string, style lipgloss.Style) (int, int) {
+	render := style.Render(s)
+	before := strings.Index(render, s)
+	after := len(render) - len(s) - before
+	return before, after
 }
 
 // UniqueStrings returns a list of unique strings given a list of strings.
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -1,0 +1,24 @@
+package utils
+
+import "github.com/charmbracelet/lipgloss"
+
+// LipglossTruncate truncates a given line based on its lipgloss width.
+func LipglossTruncate(s string, width int) string {
+	var i int
+	for i = 0; i < len(s) && lipgloss.Width(s[:i]) < width; i++ {
+	} //revive:disable-line:empty-block
+	return s[:i]
+}
+
+// UniqueStrings returns a list of unique strings given a list of strings.
+func UniqueStrings(strings []string) []string {
+	keys := make(map[string]bool)
+	var list []string
+	for _, s := range strings {
+		if _, uniq := keys[s]; !uniq {
+			keys[s] = true
+			list = append(list, s)
+		}
+	}
+	return list
+}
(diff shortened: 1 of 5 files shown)


MikaelFangel:
Current state:
demo

MikaelFangel:
@maaslalani this adds the functionality for search in the pager using regex. I hope that you agree. (:

maaslalani:
So cool! Thank you @MikaelFangel, I love that it can be given regex!

maaslalani:
Hey @MikaelFangel, search works really well, I think there are a couple of pieces of feedback that we'll need to do to get this merged in.

1. There should be some differentiation between the active search result and the inactive search results. `match` vs `match-active`
2. Search (when active) should replace the help view so that it is on the same level.
3. <kbd>N</kbd> should also be a keybind for going to the previous match.

MikaelFangel:
Thank you for the feedback! I'll give it a try during this week or the next 😊

internal/utils/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/charmbracelet/lipgloss"
+)
+
+// LipglossTruncate truncates a given line based on its lipgloss width.
+func LipglossTruncate(s string, width int) string {
+	var i int
+	for i = 0; i < len(s) && lipgloss.Width(s[:i]) < width; i++ {
+	} //revive:disable-line:empty-block
+	return s[:i]
+}
```
maaslalani, line 15: I believe you can use `runewidth.Truncate` here.

MikaelFangel: I had some issues where when using the runewidth.Truncate that it didn't return the correct results because it doesn't calculate the width correctly. 😊 But maybe there is another way of fixing this? 

As I remember it was when adding the formatting to the text that I got a bug. I can try to give a reproducible example? 😇

Also if not mistaken lipgloss actually specifies that you should use lipgloss.Width instead of len(string) and runewidth.

MikaelFangel:
@maaslalani I've made the requested changes. It has added some complexity and performance overhead, and some of it could without a doubt be improved.

maaslalani:
Hey @MikaelFangel this is really awesome work.

I'm seeing one bug where when searching, and then pressing <kbd>n</kbd> you can see a state where two matches are highlighted:

image

How to reproduce:
1. Search `shell` in gum README (I was able to reproduce with other words as well)
2. Press `n` repeatedly 6 times
3. On the 6th time the cursor doesn't highlight the next match
4. Press `n` again
5. On the 7th press both matches are highlighted

MikaelFangel:
@maaslalani - Found some other bugs while fixing the one you described and have tried to simplify the logic while doing so. But it should work now (:

maaslalani:
Awesome @MikaelFangel, just tested it out and it feels a lot more robust now!

One thing I'd request is that the help should only show `Next Match` / `Previous Match` if there is currently an active search query.

maaslalani:
@MikaelFangel Another thing I am noticing is that the calculations do not work correctly for non-softwrapped text.

If I try to search for `brew` in the Gum `README.md` normally the calculation is off by (I believe) the number of soft-wrapped lines. With `--soft-wrap` turned on the search works well.

MikaelFangel:
@maaslalani Think I found the calculation error and should have corrected it now.

maaslalani:
Incredible work @MikaelFangel, thank you so much for this PR and your patience in the review process.

maaslalani (approved):

GreatestOfAllTimee:
I wonder if this is only apply for pager? It would be great if this features can be use in filter mood.

cristiand391:
Just saw this feature in the release notes, love it!
Awesome work @MikaelFangel!

MikaelFangel:
@cristiand391 Thank you very much!

Issue #343, opened 2023-04-28
Crash in `gum file` in empty folder


johan-st:
Navigating to an empty folder (or a folder with only hidden files when these are not shown) and pressing Enter causes a crash
I have only tested this on Debian 11 on WSL2

**To Reproduce**
1. Create an empty dir
2. navigate into it with `gum file`
3. press enter when in the folder

**Expected behavior**
I expect nothing to happen when hitting in this senario

**Screenshots**
![VHS of crash](https://vhs.charm.sh/vhs-1PU2L3iC47AwMMVZ9VoD0.gif)

OS: Debian 11 on WSL2

maaslalani:
Fixed by:
* https://github.com/charmbracelet/bubbles/pull/374
* https://github.com/charmbracelet/gum/commit/99a1fa9d8e8af227e6a7eafcd6c35067941b5369

PR #348, opened 2023-05-01
fix: missing filepicker styles

caarlos0:
some styles were not being set, for instance, the "no items" style

refs #343


maaslalani:
Thanks so much! ❤️

PR #357, opened 2023-05-18
fix: measure runes instead of bytes when truncating

meowgorithm:
I noticed a couple of gotchas in `LipGlossTruncate`:

* `len()` returns the number of bytes in a string
* Slicing a string slices it on a byte level

The fix would normally be to convert the string to a slice of runes and operate on that new slice:

```go
r := []rune(str)
```

However, reflow already contains an ansi-aware truncate feature that we can use instead as a drop-in replacement for the LipGlossTruncate function.

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -6,14 +6,6 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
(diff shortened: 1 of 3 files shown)


maaslalani:
Thanks so much!

PR #371, opened 2023-06-08
Add option to show output on error

### LINKED ISSUE #55: [feature request] add option for `gum spin` to dump output on error
I think it would be great if `gum spin` could act as a prettier version of [chronic](https://manpages.debian.org/jessie/moreutils/chronic.1.en.html), which hides the output *unless* an error occurs

jelle-bigbridge:
Fixes #55

### Changes
- Add option to show output on error


westhomas:
This is _really_ needed for when commands fail. Otherwise there's no good error messages from stderr and your users have to re-run the command by hand trying to debug why it failed.

spin/options.go
```diff
@@ -10,11 +10,12 @@ import (
...
 
-	ShowOutput   bool          `help:"Show or pipe output of command during execution" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
-	Spinner      string        `help:"Spinner type" short:"s" type:"spinner" enum:"line,dot,minidot,jump,pulse,points,globe,moon,monkey,meter,hamburger" default:"dot" env:"GUM_SPIN_SPINNER"`
-	SpinnerStyle style.Styles  `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
-	Title        string        `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
-	TitleStyle   style.Styles  `embed:"" prefix:"title." envprefix:"GUM_SPIN_TITLE_"`
-	Align        string        `help:"Alignment of spinner with regard to the title" short:"a" type:"align" enum:"left,right" default:"left" env:"GUM_SPIN_ALIGN"`
-	Timeout      time.Duration `help:"Timeout until spin command aborts" default:"0" env:"GUM_SPIN_TIMEOUT"`
+	ShowOutput    bool         `help:"Show output of command" default:"false" env:"GUM_SPIN_SHOW_OUTPUT"`
+	OutputOnError bool         `help:"Show output of command if it fails" default:"false" env:"GUM_SPIN_OUTPUT_ON_ERROR"`
```
maaslalani, line 14:
```suggestion
	ShowError bool         `help:"Show output (error) of command if it fails" default:"false" env:"GUM_SPIN_SHOW_ERROR"`
```

spin/options.go
```diff
@@ -10,11 +10,12 @@ import (
...
+	Spinner       string       `help:"Spinner type" short:"s" type:"spinner" enum:"line,dot,minidot,jump,pulse,points,globe,moon,monkey,meter,hamburger" default:"dot" env:"GUM_SPIN_SPINNER"`
+	SpinnerStyle  style.Styles `embed:"" prefix:"spinner." set:"defaultForeground=212" envprefix:"GUM_SPIN_SPINNER_"`
+	Title         string       `help:"Text to display to user while spinning" default:"Loading..." env:"GUM_SPIN_TITLE"`
+	TitleStyle    style.Styles `embed:"" prefix:"title." envprefix:"GUM_SPIN_TITLE_"`
+	Align         string       `help:"Alignment of spinner with regard to the title" short:"a" type:"align" enum:"left,right" default:"left" env:"GUM_SPIN_ALIGN"`
+  Timeout      time.Duration `help:"Timeout until spin command aborts" default:"0" env:"GUM_SPIN_TIMEOUT"`
```
maaslalani, line 20:
```suggestion
    Timeout      time.Duration `help:"Timeout until spin command aborts" default:"0" env:"GUM_SPIN_TIMEOUT"`
```

maaslalani:
Hey, thanks for the contribution. Can we call this `ShowError` instead of `OutputOnError` to match a little better with `ShowOutput`.

jelle-bigbridge:
@hopefulTex could you take a look at this PR. You changed some stuff that this PR depends on, and I'm not sure how to do it now. I think this would still be a really useful feature so maybe you could help make this PR work

hopefulTex:
Pardon the dust!

Adding a distinction is definitely useful, I'd love to get this updated soon but a few questions still remain.

Currently spin is supposed to show both the output and the error in one block, with the ui being displayed in the `stderr`, with both outputs mirrored to `stdout` when using the `--show-output` flag.

- Should the error be put on the `stderr` or the `stdout`?
 - If we use `stderr`, how do we distinguish between the ui and the actual error?
 - If `stdout` is used for the error, how can this play nicely with the `--show-output` flag?
- Is this feature meant for parsing errors with a script or program, or will a human-readable separator suffice?
 - How can we make a separator that plays nice with screen-readers?

I'll be updating this to match with the current version of gum, but I believe these questions are probably better answered by people who want/need this for their script or workflow (not me).

jelle-bigbridge:
My preferred implementation (I want to use it in a script, and I was using my previous implementation in a script already). Is that stdout and stderr are combined by the system (idk how this works in go but the unix system call is dup2), so that it is the same as if running the command on the shell, and messages on stdout and stderr are interleaved.

Only when the subcommand exits with a non-0 exit code, will the output be shown to the user


> * Should the error be put on the `stderr` or the `stdout`?

stderr seems like it makes sense

> * If we use `stderr`, how do we distinguish between the ui and the actual error?

no distinction is needed, as it should read the same as if the command was normally executed in a user shell session

> * If `stdout` is used for the error, how can this play nicely with the `--show-output` flag?

I don't think it makes sense to use the --show-output flag in combination with this feature

> * Is this feature meant for parsing errors with a script or program, or will a human-readable separator suffice?

meant for humans

> * How can we make a separator that plays nice with screen-readers?

n/a

maaslalani:
Hey @jelle-bigbridge, thank you so much for the PR. `gum spin` added some new changes for live streaming the output of `gum spin` which would require some changes to this PR.

A PR has been opened to fix this `--show-error` issue which accounts for the new changes: https://github.com/charmbracelet/gum/pull/440

jelle-bigbridge:
Let's close this in favour of #440

PR #374, opened 2023-06-19
feat(filter): add cursor text line styling

vahnrr:
Fixes nothing, this is a feature I wanted to have access to in `gum`.

### Changes

Before, only `GUM_FILTER_TEXT_*` could be styled, this meant every line's text would look the same, even the one with the cursor on:
pre

This PR provides a way to style `gum filter`'s cursor text through `GUM_FILTER_CURSOR_TEXT_*` environment variables as such:
post

I tried to stick to your naming conventions, feel free to rename `CURSOR_TEXT` if you feel like it doesn't match the rest of the project.

Thanks for making `gum`, I really enjoy the added interactivity it brings to shell scripts!


maaslalani:
Thank you so much for this PR! I really appreciate it.

PR #380, opened 2023-06-29
feat: Adding timeout option to Filter command

deicon:
Fixes #...

### Changes
Using Timeout Option on Filter Command


maaslalani:
Thanks @deicon!

PR #384, opened 2023-06-30
feat: Timeout for Choose Command

deicon:
Fixes #...

### Changes
Timeout for Choose Command

### DIFF
--- a/choose/choose.go
+++ b/choose/choose.go
@@ -12,6 +12,9 @@ package choose
 
 import (
 	"strings"
+	"time"
+
+	"github.com/charmbracelet/gum/timeout"
 
 	"github.com/charmbracelet/bubbles/paginator"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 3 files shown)


maaslalani:
For this `choose` command I wonder if we should have the timeout simply return no values rather than what is currently selected. I'm open to both options (returning the selected and not returning anything) but leaning towards "timing out" as an error to avoid unexpected stuff.

deicon:
I think this was discussed on discord or here somewhere. Don't recall really what the outcome was. But looking at it now, I would argue, that the choose is actually a good fit for timing out and have it select a default. But as you, I would agree to both versions

maaslalani:
Hmmm actually yeah, I think you're right. If we assume the timeout would be long enough that any human would've reasonably selected their option then it definitely makes sense to return the first option as a default assuming that the user hasn't moved the cursor.

maaslalani:
Thank you so much for all your work on this feature!!!

deicon:
I am here to learn and I really like the tools you are creating here. So It was a pleasure to do this. Only the time to review could become better. :) But I understand that there is a lot on your plates. So no worries.

Issue #325, opened 2023-03-29
The color is missing when redirecting stderr to stdout since v0.10.0


Romain7495:
**Describe the bug**
The color is missing when redirecting stderr to stdout

**To Reproduce**
```gum confirm "Are you sure to perform this action ?" 2>&1 | tee```
or 
```gum confirm "Are you sure to perform this action ?" |& tee```

**Expected behavior**
Having the color even if we redirect stderr to stdout

**Screenshots**
Not working since `v0.10`
![image](https://user-images.githubusercontent.com/15386726/228475851-22a420ec-6704-4670-8511-d61ec8a005fb.png)

Working without redirect
![image](https://user-images.githubusercontent.com/15386726/228476131-97e2d8af-4d5c-46ce-b6f0-adeb66255abe.png)

**Desktop**
 - ALL platform
 - wsl (ubuntu) , linux, macos

bagl3y:
Same behavior here with 0.10.0 on WSL2 v5.15.90.1 kernel , no issues with 0.9.0

hopefulTex:
Does this issue still occur?
Testing post- #324 seems to show no problems.

dannyloxavier:
@hopefulTex the latest release still 0.10.0 to me.
I download the binary.
I confirm it still occur.

hopefulTex:
It seems that since `termenv` is detecting the pipe, it's getting rid of the ansi sequences to avoid spitting out trash to `tee`.
There are ~two ways to fix this immediately,
Technically, you should be able to force a profile by setting the `CLICOLOR_FORCE` environment variable, however it did not work on my system.
If you're okay with running your own fork you can change line 31 from
`lipgloss.SetColorProfile(termenv.NewOutput(os.Stderr).Profile)`
to
`lipgloss.SetColorProfile(termenv.ANSI256)`, with `ANSI256` replaced by whatever suits your setup. [Palettes can be found here](https://github.com/muesli/termenv/blob/20f96565abe82d2c12792808f0d6bdba373e4141/profile.go) :)

If you put in a feature request for a 'force color' option, I won't be saddled by exams by mid-May if someone else doesn't get to it first :)

vtmx:
In my case I change of colors of elements:

 ```
 export GUM_CONFIRM_UNSELECTED_BACKGROUND="0"
 export GUM_CONFIRM_UNSELECTED_FOREGROUND="15"
 export GUM_CONFIRM_SELECTED_BACKGROUND="15"
[...]

PR #387, opened 2023-06-30
Switch to termenv.EnvColorProfile()

kennyp:
This will allow users to set `NO_COLOR` and `CLICOLOR_FORCE` when they need to override the detected values.

### DIFF
--- a/format/formats.go
+++ b/format/formats.go
@@ -54,7 +54,7 @@ func markdown(input string, theme string) (string, error) {
 }
 
 func template(input string) (string, error) {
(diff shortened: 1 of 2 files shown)


kennyp:
A fix for #325 and #395. It also allows me to write a Python wrapper that works as expected when doing `format`, `style`, and friends.

kennyp:
Additionally, it would probably be good to cut a new release, as there have been several commits since v0.10.0

muesli:
Looks good to me!

caarlos0 (approved):

maaslalani:
Thanks so much @kennyp!

PR #398, opened 2023-07-17
Add `--select-if-one` flag to `choose`/`filter`.

### LINKED ISSUE #396: gum choose: Option to not ask when there is no choice
When using dynamic choice list in `gum choose`, sometimes it's no need to ask user for choice if there's just one option. So in that case it would be useful to have option to skip asking user and just return that single value.

Like `gum choose --no-ask-if-no-choice one` would just return `one` immediately.

Well alternative is to script `gum` call, but such option would be easier for users.

**Additional context**
It's also important that this should not be enabled by default, because user may expect interactive step and may press `enter` in advance and if step is skipped wrong choice could be selected,

kennyp:
Fixes #396 

### Changes
- Add `--select-if-one` to `choose`, returns the only option if only one option is given
- Add `--select-if-one` to `filter`, returns the only option if only one option is given

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -35,6 +35,11 @@ func (o Options) Run() error {
 		o.Options = strings.Split(strings.TrimSuffix(input, "\n"), "\n")
 	}
 
+	if o.SelectIfOne && len(o.Options) == 1 {
+		print(o.Options[0])
+		return nil
+	}
+
 	// We don't need to display prefixes if we are only picking one option.
 	// Simply displaying the cursor is enough.
 	if o.Limit == 1 && !o.NoLimit {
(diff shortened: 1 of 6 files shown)


kennyp:
@maaslalani and @bashbunni what's y'all's preferred way of working? Do y'all like PRs like this or would you prefer discussion in discord or something like that?

I love the stuff y'all are putting out and I would love to contribute more, so I just want to make sure my approach matches what y'all are looking for.

maaslalani:
Thanks so much for the PR @kennyp! This looks great, thank you for the contribution ❤️

vt-alt:
Thank you!

PR #404, opened 2023-07-27
spin: fix isTTY check

gregkh:
The isTTY check is inverted in the --show-output option for spin, so no output is shown anymore.  Fix this by correctly checking if we are a tty or not.


hopefulTex:
This check is meant for when the output is being piped to another program, as the `spin` ui uses `stderr` if it's in a TTY.
The output fails to display because the `--show-output` flag is not passed to the program model as of patch #385

Issue #320, opened 2023-03-22
Gum Table - Add --return-column & --show-columns


CaelFrost:
**Is your feature request related to a problem? Please describe.**
Rematching data from the selected row to the source table, so that once a choice is made the script can use the full row data versus the subset shown in gum table. 

**Describe the solution you'd like**

- When sending data in as a table, I often want to include an index column but not show that column to the user in the TUI. 
     not sure if the right choice is to use --show-columns or --hide-columns, but either would solve my need
- Once the user makes the selection, having gum table return ONLY the columns specified in --return-column=0 will cause gum table to only return the first column of each row, and allows the script to use this returned data to refilter the source table for only the indexes specified.

**Describe alternatives you've considered**
currently adding a 'index' column but this is shown to the user

**Additional context**

deicon:
#415 Should fix the request for the return value

deicon:
I prefer to use 1 if referring to the first column in user facing command args.

deicon:
https://github.com/charmbracelet/bubbles/pull/412 Will be needed to support the hiding feature. 
Once and If this will be merged to bubbles, I will create a PR here for the --hide-columns feature

caarlos0:
we could probably change the table after parsing the CSV and remove the columns we don't want, or something like that...

deicon:
The return column should be possible now. Hiding columns should also be possible by setting width to 0?

caarlos0:
yes, you're right, columns can be hidden by providing `--widths=10,0,10` (to hide the 2nd column for example).

thanks @deicon , will close this!

PR #415, opened 2023-08-23
feat: Adding Return Column to table command

deicon:
Consider a table of
ID Name Description
1  Task1 Task description

It would be good to select the row but retrieve the ID for example for further processing

like 'task 1 delete' when first row was selected

### Changes
- table command. Adding extra parameter to select column which will be returned on selection


deicon:
Fixes #320

deicon:
@maaslalani Is there still demand for this kind of feature ?

table/options.go
```diff
@@ -16,4 +16,6 @@ type Options struct {
 	CellStyle     style.Styles `embed:"" prefix:"cell." envprefix:"GUM_TABLE_CELL_"`
 	HeaderStyle   style.Styles `embed:"" prefix:"header." envprefix:"GUM_TABLE_HEADER_"`
 	SelectedStyle style.Styles `embed:"" prefix:"selected." set:"defaultForeground=212" envprefix:"GUM_TABLE_SELECTED_"`
+	File          string       `short:"f" help:"file path" default:""`
```
caarlos0, line 19:
```suggestion
```

caarlos0 (approved):

PR #424, opened 2023-09-17
fix(filter): made filter work with lists as choose

### LINKED ISSUE #399: Why filter doesn't work with lists as choose does?
**The Bug**
`gum choose Apple Banana Orange` --> Works
`gum filter Apple Banana Orange` --> Doesn't

**Expected behavior**
Filter too should accept lists as input like Choose.

MikaelFangel:
Fixes #399

### Changes
- Adds the options parameter to filter to let kong parse correctly


filter/command.go
```diff
@@ -34,18 +33,16 @@ func (o Options) Run() error {
...
+			o.Options = files.List()
 		}
-	} else {
-		choices = files.List()
 	}
 
-	if len(choices) == 0 {
-		return errors.New("no options provided, see `gum filter --help`")
-	}
+	choices = make([]string, len(o.Options))
```
bashbunni, line 44: We already declared a `choices` var on line 35, any reason to re-declare here? Also why assign input values to `o.Options` instead of using `choices`? Just want to make sure I'm not missing anything :)

filter/command.go
```diff
@@ -34,18 +33,16 @@ func (o Options) Run() error {
 	v := viewport.New(o.Width, o.Height)
 
 	var choices []string
-	if input, _ := stdin.Read(); input != "" {
-		input = strings.TrimSuffix(input, "\n")
-		if input != "" {
-			choices = strings.Split(input, "\n")
+	if len(o.Options) == 0 {
+		if input, _ := stdin.Read(); input != "" {
```
bashbunni, line 37: Also here I would lean towards using a positive conditional statement so it's more readable. "If the input is empty, we do this, otherwise we use its value"

bashbunni:
While we're at it, if you would like to provide some tests to make sure the filtering is working as expected that would be great. Tests for `exactMatches` and `matchAll` would be super helpful. Only if you're able, otherwise I can create an issue for new contributors/hacktoberfest :)

MikaelFangel:
Thanks for the review. I'll try to take at look at the review this evening 😊

bashbunni:
@MikaelFangel No rush, just whenever works :) Thanks again for the hard work!

filter/command.go
MikaelFangel: Actually, when looking at it, it doesn't make sense to introduce the variable choices here. I'm changing the rest of `filter/command.go` to o.Options because o.Options is already declared and used by kong.

filter/command.go
MikaelFangel: Here o.Options is already an input, and it follows the same logic as `choose/command.go`. So I think either it should be changed in both places or kept as is, so there is a greater similarity in the structure of the two modules.

filter/command.go
```diff
@@ -34,18 +33,16 @@ func (o Options) Run() error {
...
+			o.Options = strings.Split(strings.TrimSuffix(input, "\n"), "\n")
+		} else {
+			o.Options = files.List()
 		}
-	} else {
-		choices = files.List()
 	}
 
-	if len(choices) == 0 {
```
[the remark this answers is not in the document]
MikaelFangel: I'll add it back (:

MikaelFangel:
@bashbunni I've tried to implement your suggestions from the code review except the one with the positive conditional statement. Let me know if it should be changed any further.

I'd happily contribute to gum any day. (:

maaslalani:
Thank you so much for the changes @MikaelFangel! We really appreciate all your contributions ❤️

bashbunni:
Awesome, thank you! @MikaelFangel

PR #427, opened 2023-09-18
feat(confirm): add `--show-output`

vahnrr:
Fixes nothing, this is a feature I found missing from a specific use case I had in mind.

### Changes

Adds the `--show-output` flag to `gum confirm` which prints the prompt and the chosen action to STDOUT.

```shell
gum confirm --show-output 'Do you agree?'
# Do you agree? Yes

gum confirm --show-output 'Do you agree?'
# Do you agree? No
# exit status 1
``` 

This merge adds 2 commits, only the first one (`4f9235b`) is required for the feature to work.
Since the feature doesn't affect `confirm/confirm.go` I feel like the second (`432182f`) isn't necessary, but I still included it in case I misunderstood the purpose of the model structure.

### DIFF
--- a/confirm/command.go
+++ b/confirm/command.go
@@ -19,6 +19,7 @@ func (o Options) Run() error {
 	m := model{
 		affirmative:      o.Affirmative,
 		negative:         o.Negative,
+		showOutput:       o.ShowOutput,
 		confirmation:     o.Default,
 		defaultSelection: o.Default,
 		keys:             defaultKeymap(o.Affirmative, o.Negative),
--- a/confirm/command.go
+++ b/confirm/command.go
@@ -38,6 +38,14 @@ func (o Options) Run() error {
 		return fmt.Errorf("unable to confirm: %w", err)
 	}
 
+	if o.ShowOutput {
+		confirmationText := m.negative
+		if m.confirmation {
+			confirmationText = m.affirmative
+		}
+		fmt.Println(m.prompt, confirmationText)
+	}
+
 	m = tm.(model)
 	if m.confirmation {
 		return nil
--- a/confirm/confirm.go
+++ b/confirm/confirm.go
@@ -82,6 +82,7 @@ type model struct {
 	help        help.Model
 	keys        keymap
 
+	showOutput   bool
 	confirmation bool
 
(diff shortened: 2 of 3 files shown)


confirm/command.go
```diff
@@ -34,10 +35,21 @@ func (o Options) Run() error {
 
 	if m.(model).aborted {
 		os.Exit(exit.StatusAborted)
-	} else if m.(model).confirmation {
-		os.Exit(0)
+
```
MikaelFangel, line 38:
```suggestion
```

confirm/command.go
```diff
@@ -34,10 +35,21 @@ func (o Options) Run() error {
...
+			}
+			fmt.Println(m.(model).prompt, confirmationText)
+		}
+
+		if m.(model).confirmation {
+			os.Exit(0)
+		} else {
+			os.Exit(1)
+		}
 	}
```
MikaelFangel, line 53: Here you should remove the else and outdent the code and also the last else statement can be outdented as well.

The code will look something like this.
```
	}

	if o.ShowOutput {
		confirmationText := m.(model).negative
		if m.(model).confirmation {
			confirmationText = m.(model).affirmative
		}
		fmt.Println(m.(model).prompt, confirmationText)
	}

	if m.(model).confirmation {
		os.Exit(0)
	}

	os.Exit(1)
```

MikaelFangel (changes requested):
I have added some suggestions to help fix your linting error.

Also, you can check if it passes linting before pushing by using `golangci-lint run`

confirm/command.go
vahnrr: I must have missed the CI result notifications, really sorry.
This should be fixed as per you suggestions, thanks!

Regarding the model structure, do you think [this commit](https://github.com/charmbracelet/gum/pull/427/commits/432182f4b8eefa0aa006daa8c672a15d8da4beb9) is necessary?

caarlos0:
this looks good! can you fix the conflicts? if not, I can re-implement it on main...

vahnrr:
No problem, it should be up to date now

caarlos0:
thanks @vahnrr

Issue #286, opened 2023-02-25
In gum table, escape key does not quit


hydrargyrum:
**Describe the bug**
In gum table, escape key does not quit

**To Reproduce**
1. run `gum table <<< "a,b" <<< "1111111111111,2222222222"`
2. press escape
3. gum did not exit :(

**Expected behavior**
gum table exits with an error, just like `gum input` or `gum choose`

**Desktop (please complete the following information):**
 - OS: debian
 - Version https://github.com/charmbracelet/gum/releases/tag/v0.9.0

maaslalani:
Fixed by #290

PR #433, opened 2023-09-28
feat(write): ESC should be successful

docwhat:
Change ESC from aborting to successful quitting.

'vi' users press ESC as an uncontrollable tick, making using 'gum write' painful when all their work is lost.

I'm floating this as an idea.

I could add a non-zero exit code, mostly preserving the current behavior.  If I did that, ctrl-c should also return the text.

If we standardized the exit code, we could apply it to #286.

### Changes

- ESC now indicates a normal exit (like pressing CTRL+D).

### DIFF
--- a/write/write.go
+++ b/write/write.go
@@ -46,11 +46,11 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 	case tea.KeyMsg:
 		switch msg.String() {
-		case "ctrl+c", "esc":
+		case "ctrl+c":
 			m.aborted = true
 			m.quitting = true
 			return m, tea.Quit
-		case "ctrl+d":
+		case "ctrl+d", "esc":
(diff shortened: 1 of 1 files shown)


maaslalani:
I think I agree with this for the most part and think people would also benefit from this. If it's complained about we will probably revert though since it is a slightly breaking change.

docwhat:
> If it's complained about we will probably revert though since it is a slightly breaking change.

Just thinking out loud:

- The end user (script-user) is the one who should have precedence here. They are the ones who the change impacts.
- The developer (script-writer) may have a preference for a default.
- The developer may have a legitimate reason to override the user. Even if I can't think of one at the moment.
- There may be a well-known place to pick up a default.
    - Maybe checking the shell for vi mode? 
    - See if `$EDITOR` is set to a `vi` clone?
    - Is this something `~/.readline` might have?

PR #440, opened 2023-10-14
feat(spin): Add support for `--show-error` for the spinner.

### LINKED ISSUE #55: [feature request] add option for `gum spin` to dump output on error

elliotcourant:
This makes it so that if the `--show-error` flag is provided then the
full output of the command will be printed if the command fails. This
kind of works in conjuncture with `--show-output` in that if the command
succeeds only STDOUT is pushed. If the command fails both `STDOUT` and
`STDERR` are pushed.

This builds off of https://github.com/charmbracelet/gum/pull/371

Resolves #55

Fixes #55

### Changes
- Adds support for a `--show-error` flag for the spinner command.

---

### Examples

```bash
gum spin -s minidot --show-error -- bash -c "sleep 1 && echo 'you can see me if the command fails' && exit 1"
# you can see me if the command fails
# exit status 1
```

```bash
gum spin -s minidot --show-error -- bash -c "sleep 1 && echo 'you can see me if the command fails' && exit 0"
# (no output)
```

### DIFF
--- a/spin/spin.go
+++ b/spin/spin.go
@@ -15,6 +15,7 @@
 package spin
 
 import (
+	"io"
 	"os/exec"
 	"strings"
 	"time"
@@ -34,16 +35,22 @@ type model struct {
 	aborted    bool
 	status     int
 	stdout     string
+	stderr     string
+	output     string
 	showOutput bool
+	showError  bool
 	timeout    time.Duration
 	hasTimeout bool
 }
 
+var bothbuf strings.Builder
 var outbuf strings.Builder
 var errbuf strings.Builder
 
 type finishCommandMsg struct {
 	stdout string
+	stderr string
+	output string
 	status int
 }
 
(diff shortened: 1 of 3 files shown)


spin/spin.go
```diff
@@ -55,8 +62,11 @@ func commandStart(command []string) tea.Cmd {
 		}
 		cmd := exec.Command(command[0], args...) //nolint:gosec
 
-		cmd.Stdout = &outbuf
-		cmd.Stderr = &errbuf
+		stdout := io.MultiWriter(&bothbuf, &errbuf)
+		stderr := io.MultiWriter(&bothbuf, &outbuf)
```
elliotcourant, line 66: By doing this the `bothbuf` will have the output of both STDERR and STDOUT but it will be in the order it was written by the process. 

If we have the buffers entirely separate and the writes to each were interleaved in any way, this might cause the output to be confusing or be inaccurate.

pingiun:
I would really like to have this feature in gum, can I help to get it merged?

elliotcourant:
This PR has been taken over in https://github.com/charmbracelet/gum/pull/440

PR #452, opened 2023-11-08
Log predefined time formats

maaslalani:
Add predefined time formats

### DIFF
--- a/log/command.go
+++ b/log/command.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"os"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


aymanbagabas (approved):
Any way we can display that in the help?

caarlos0 (approved):

caarlos0:
> Any way we can display that in the help?

on completions would be nice too

maaslalani:
> Any way we can display that in the help?

We can do it if we make it an enum, which will prevent any custom time formats though, which may be okay since the predefined list is quite exhaustive.

PR #457, opened 2023-11-28
Hide Style Flags consistently

maaslalani:
Prevents style flags from leaking through when a user makes an error in the usage.

For example:

```
gum write --help
```

using the `--help` flag produces a different usage than the user making an error with the flags.

```
gum write --error
```

This is because we perform the hiding of the flags on `BeforeReset` which is called for the help flag but not on an error.

The downside of this approach is that we must duplicate the style struct so that we can maintain a fully hidden embeddable struct while also having a non-hidden struct so the style command has proper usage / help / errors.

This can be resolve once the following issue is implemented:

https://github.com/alecthomas/kong/issues/316

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,8 +48,9 @@ func main() {
 		kong.Description(fmt.Sprintf("A tool for %s shell scripts.", bubbleGumPink.Render("glamorous"))),
 		kong.UsageOnError(),
 		kong.ConfigureHelp(kong.HelpOptions{
-			Compact: true,
-			Summary: false,
+			Compact:             true,
+			Summary:             false,
+			NoExpandSubcommands: true,
 		}),
 		kong.Vars{
 			"version":                 version,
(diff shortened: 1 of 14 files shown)


aymanbagabas (commented):
Instead of defining two style structs, why don't we mark the flags hidden by default and "unhide" them during the `BeforeReset` hook and when encountering a `--help` flag?

aymanbagabas (approved):
LGTM, I would add a comment referring to the issue url in the code

main.go
```diff
@@ -74,7 +75,6 @@ func main() {
 		if errors.Is(err, exit.ErrAborted) {
 			os.Exit(exit.StatusAborted)
 		}
-		fmt.Println(err)
```
acidghost, line 77: @maaslalani  Why was this deleted? Now it's not printing any error.

PR #463, opened 2023-12-10
fix: Make --select-if-one print to stdout

snan:
For some reason it wasn't printing to stdout (and I could repro that
bug even on versions before I added the newline). It was only showing
up on other streams in the shell (error stream probably), not getting
sent into pipes.

I changed it to fmt.Println.

As for the ansi-stripping that was in `filter`, LMK if that's what you
prefer and I'll add it to `choose` too. I just wanted them to match.


maaslalani:
Thank you so much! ❤️

PR #474, opened 2023-12-20
docs: add openSUSE installation instructions for gum

thiagovictorino:
This commit introduces installation instructions for the gum tool in the README.md, tailored explicitly for openSUSE users. It includes steps for adding the Charm repository and installing gum using Zypper. This update ensures that openSUSE users have clear guidance on installing gum, aligning with the existing documentation for other operating systems.

### Changes
- Add documentation in the `README.md` to include how to install using zypper


caarlos0:
maybe we can merge it with the other rpm instructions? the repo setup part is the same, so it'll save some real estate

PR #522, opened 2024-03-28
Use Huh for Gum Confirm

maaslalani:
Use `huh` for `gum confirm`.


farzadmf:
@maaslalani seems like this broke the [shift-]tab bindings to navigate the choices and also pressing `y` or `n` to accept the choices; any plan to add those?

PR #534, opened 2024-04-07
docs: add install instructions to Exherbo Linux in README.md

davlgd:


caarlos0:
I think we're going to keep the most popular/well-known distros in the readme for now.

people can easily find this PR though through search though, which is great!

thanks for letting us know this distro exists, by the way :D 

cheers!

PR #536, opened 2024-04-11
Docs: fixed gum sudo replacement

olivertzeng:
only for zsh


maaslalani:
Hey! Thanks for this PR. I am going to close this for now as I don't want to overwhelm users when they are reading the README. This example is just for them to get a sense of what they can do with gum. But we can add this more complex script in an example if you're up for that!

olivertzeng:
Alright!I will surely open another one for that!
 Sent from Proton Mail for iOS 
於 週日, 5月 26, 2024 時間 02:45, Maas Lalani ***@***.***&gt; 寫道：  
Hey! Thanks for this PR. I am going to close this for now as I don't want to overwhelm users when they are reading the README. This example is just for them to get a sense of what they can do with gum. But we can add this more complex script in an example if you're up for that!

Issue #393, opened 2023-07-09
Add more intuitive key bindings to gum choose like pagers (such as less) have


vt-alt:
**Is your feature request related to a problem? Please describe.**
`choose` key bindings aren't intuitive for a Linux user. When presented with visual choose user expects navigations keys to work (as in pager, like they works in `less`) but `Home`/`End` key don't work. Also, interface does not present itself like it's `vim` (so even if `gum` user is a `vim` user they don't guess to use `vim` keys). I think `Home`/`End` is more intuitive than `g`/`G`. So for occasional `gum` user key bindings aren't intuitive and intuitive keys aren't used.

**Describe the solution you'd like**
Please add `Home`/`End` keys support for scrolling to the top/bottom in choose mode. And (less important add)  `q` to exit `choose`. 🙏

Home/End/q is what Linux user expects to work from a pager and these keys seems not used already in `choose`.

**Describe alternatives you've considered**
Maybe better is to add configuration option (like env variable) to add different binding, but this is more complicated than to add default support, and I think by default support for Home/End/q is better, as _pager_ behavior is more intuitive.

vt-alt:
@maaslalani Great! thx

PR #548, opened 2024-04-26
Support using the Home/End keys in pager

lzm0:
Since `choose` also supports Home/End for navigation in https://github.com/charmbracelet/gum/issues/393.
It would be nice if `pager` can do the same.


maaslalani:
Thanks so much @lzm0! Appreciate the PR ❤️

PR #576, opened 2024-05-12
chore: update nix

MikaelFangel:
I noticed that all nix related files are 2 years old, so I've created this PR to address this issue. Otherwise I would suggest to remove the files and refer to Nixpkgs instead. (: 

### Changes
- Changed Nixpkgs to the unstable branch to avoid having to update at each release.
- Updated the flake.lock accordingly.
- Add the ldflags to the default.nix file.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,7 +1,12 @@
 { pkgs }:
 
-pkgs.buildGoModule {
-  name = "gum";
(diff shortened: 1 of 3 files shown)


maaslalani:
Awesome work!

PR #577, opened 2024-05-12
fix(stdin): add trim of trailing newline

### LINKED ISSUE #573: Unhandled trailing newline from stdin in `gum input`
**Describe the bug**

`echo foo | gum input` gives you a input prompt where the starting value has trailing whitespace.

**To Reproduce**

* trailing whitespace: `echo foo | gum input`
* no trailing whitespace: `gum input --value="$(echo foo)"`

**Expected behavior**

The final trailing newline from stdin should be filtered. The posix standard stated that all text files (and thus in turn stdout) must feature a trailing newline, and tools should respect this.

**Desktop:**
 - OS: Manjaro and NixOS

MikaelFangel:
Fixes #573

### Changes
- When receiving inputs via stdin trim the trailing newline.
- Removed trimsuffix from other commands where stdin was read.

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -22,7 +22,7 @@ func (o Options) Run() error {
 		if input == "" {
 			return errors.New("no options provided, see `gum choose --help`")
 		}
-		o.Options = strings.Split(strings.TrimSuffix(input, "\n"), "\n")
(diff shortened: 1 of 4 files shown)


maaslalani:
Awesome refactor!

PR #579, opened 2024-05-13
fix(input): make keymap backwards compatible

### LINKED ISSUE #562: "Esc" does not exit "gum input" anymore
**Describe the bug**
since 0.14.0, 'Esc' does not cancel "gum input" anymore

**To Reproduce**
gum input --placeholder "Test"

**Expected behavior**
When pressing "Esc" key, should react as when pressing "Ctrl-C", just like before.

 - OS: Windows 11
 - Powershell 7.4.2
 - Gum v0.14.0

MikaelFangel:
Fixes #562

### Changes
- Adds a custom keymap to input so the escape behavior is backwards compatible. 

I don't know if there were other keymaps that were overwritten by the 0.14.0 update, but in that case it would maybe make sense to define a keymap at a higher level and use that or even make it an option to keep compatibility.


maaslalani:
Awesome work @MikaelFangel!

PR #598, opened 2024-06-04
Add key/value feature for `gum choose` (#530)

### LINKED ISSUE #530: filter/choose: output the chosen index(s) **or** add display name support
If choices are kinda complex or verbose, I may wish to provide a display name instead just the raw value. Currently I would have to then map my display name back to the actual value.

An option flag, such as `--indexed`, which causes the selection's index to be printed, instead of its value.

```shell
$ gum choose --indexed \
     "display name one" "display name two" "display name three" "display name four"

  display name one
> display name two
  display name three
  display name four

# Prints:
1
 ```

Note: the index should refer to the supplied arguments, not necessarily the order they are displayed in. Unless `--ordered` is used.

- Allowing a list of display names to be supplied in addition to the values to return.
- Allowing a list of displayname+value pairs to be supplied.

**Additional context**
It would also be nice to show additional context on hover; perhaps if one of the alternatives is used, then the raw value could be displayed below the display-name while that option is selected?

```shell
$ gum choose \
    --preview-value \
    --display-names "display name one" "display name two" "display name three" "display name four" \
    "actual_value_1" "actual_value_2" "actual_value_3" "actual_value_4"

  display name one
> display name two
      (actual_value_2)
  display name three
  display name four

# Prints:
actual_value_2
 ```
 
 This could be awkward to work with, since you'd need to assume half the arguments are display names and the other half are values...
 
 An alternative:

```shell
$ gum choose \
    --no-preview-value \
    --deliminator "=" \
    "display name one=actual_value_1" "display name two=actual_value_2" "display name three=actual_value_3" "display name four=actual_value_4"

  display name one
> display name two
  display name three
  display name four

# Prints:
actual_value_2
 ```

This would require additional data-processing on the user's part though... At that point it probably makes more sense to use the indexed approach? Or go all-out and use JSON instead of custom deliminator.

### PR DESCRIPTION
**Pull request by clysto**
Fixes #530

### Changes
- Use huh to implement custom `gum choose` key and value.

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -39,7 +39,9 @@ func (o Options) Run() error {
 		if delimIndex == -1 {
 			options[i] = huh.NewOption(option, option)
 		} else {
-			options[i] = huh.NewOption(option[:delimIndex], option[delimIndex+1:])
+			key := strings.TrimSpace(option[:delimIndex])
+			value := strings.TrimSpace(option[delimIndex+1:])
+			options[i] = huh.NewOption(key, value)
 		}
 	}
 
--- a/choose/command.go
+++ b/choose/command.go
@@ -35,13 +35,13 @@ func (o Options) Run() error {
 	theme := huh.ThemeCharm()
 	options := make([]huh.Option[string], len(o.Options))
 	for i, option := range o.Options {
-		delimIndex := strings.Index(option, o.Deliminator)
-		if delimIndex == -1 {
-			options[i] = huh.NewOption(option, option)
-		} else {
-			key := strings.TrimSpace(option[:delimIndex])
-			value := strings.TrimSpace(option[delimIndex+1:])
+		parsed := strings.SplitN(option, o.Deliminator, 2)
+		if len(parsed) == 2 {
+			key := strings.TrimSpace(parsed[0])
+			value := strings.TrimSpace(parsed[1])
 			options[i] = huh.NewOption(key, value)
+		} else {
+			options[i] = huh.NewOption(option, option)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by maaslalani** [comment]
@clysto This is awesome work thank you! Instead of using `strings.Index` could you use `strings.Split` on the delimiter?

File: choose/command.go [hunk 1]
```diff
@@ -33,7 +33,17 @@ func (o Options) Run() error {
 	}
 
 	theme := huh.ThemeCharm()
-	options := huh.NewOptions(o.Options...)
+	options := make([]huh.Option[string], len(o.Options))
+	for i, option := range o.Options {
+		delimIndex := strings.Index(option, o.Deliminator)
```
maaslalani, line 38: Can we use `strings.Split` here (just a personal preference)

maaslalani (changes requested):

choose/command.go
clysto: Sometimes there may be two ":" in the string or no ":" in the string. I think it is easier to determine these situations using the `strings.Index`.

clysto: I have now switched to using `strings.SplitN`, maybe it looks better!

clysto:
Can this be merged?

maaslalani:
@clysto Awesome work, I just tested this and it works great! Thank you for the contribution ❤️

maaslalani (approved):

meowgorithm:
Reverting this for now as there's a typo in the help text. We'll fix and re-add after the next release.

PR #616, opened 2024-07-10
Use in-house packages

aymanbagabas:
- Use `x/ansi` to strip ansi
- Use `x/term` for `isatty` and `IsTerminal`


ansi/ansi.go
```diff
@@ -1,10 +1,8 @@
 package ansi
 
-import "regexp"
-
-var ansiEscape = regexp.MustCompile(`\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])`)
+import "github.com/charmbracelet/x/ansi"
```
ccoVeille, line 3: I would add a name to this import.

Because importing ansi in ansi package doesn't sound a good idea

ansi/ansi.go
```diff
@@ -1,10 +1,8 @@
...
 
 // Strip strips a string of any of it's ansi sequences.
 func Strip(text string) string {
```
ccoVeille, line 6: If you have only this in the package, replace all call to ansi.Strip by the x/ansi one.

Or at least add a 


```suggestion
// Deprecated: use x/ansi
 func Strip(text string) string {
```

aymanbagabas: Yes, I was thinking about dropping `gum/ansi` altogether, just not sure about the impact

CC/ @maaslalani

PR #685, opened 2024-10-08
examples/test.sh: Fixes broken table example

hxlnt:
Fixes broken portion of examples/test.sh due to incorrect path to CSV. Also adds a `--height=12` to work around current bug where height is set to 0 if not specified.


caarlos0:
this is actually a bug in the table component, fixed in https://github.com/charmbracelet/gum/pull/716

thanks for the PR though 🙏🏻

hxlnt:
The linked PR doesn't appear to fix the CSV link.

caarlos0:
ah, right!

PR #718, opened 2024-11-15
fix(confirm,choose,file,input): timeout handling

### LINKED ISSUE #684: feature request: Different exit codes for --timeout and ^C
I'd like to be able to detect whether the `gum` command timed out or if the user aborted with ctrl-c. Currently both cases exit with code 130. This will allow a script to take a default action given a timeout

caarlos0:
- some fields were not actually using the `--timeout` value
- some fields had different behavior when a timeout did occur. On this matter, it seems to me the best way forward is to specifically say it timed out, and after how long
- added exit status 124 (copied from `timeout` from coreutils) (fixes #684)

so, tl;dr: this makes `--timeout` actually work everywhere it was already added, and then on timeout they'll all print `timed out after <duration>` and then `exit 124`


main.go
```diff
@@ -73,6 +73,10 @@ func main() {
 		},
 	)
 	if err := ctx.Run(); err != nil {
+		if errors.Is(err, huh.ErrTimeout) {
+			fmt.Println(err)
```
aymanbagabas, line 77:
```suggestion
			fmt.Fprintln(os.Stderr, err)
```

aymanbagabas: Shouldn't it print to stderr?

confirm/command.go
```diff
@@ -31,11 +31,11 @@ func (o Options) Run() error {
...
 		WithShowHelp(o.ShowHelp).
 		Run()
-
 	if err != nil {
-		if !o.errIsValidTimeout(err) {
-			return fmt.Errorf("unable to run confirm: %w", err)
+		if errors.Is(err, huh.ErrTimeout) {
+			return exit.NewTimeout(o.Timeout)
 		}
+		return err
```
ccoVeille, line 38: You are repeating this exact code in multiple places in multiple files

Why not adding a function

Something like

```go
if err != nil {
   return o.handleTimeout(err)
}
```

And then define handleTimeout method to return exit.NewTimeout on huh.Timeout

ccoVeille (approved):

confirm/command.go
caarlos0: each command has its own `Options`, would have to repeat anyway.

thought about moving it to `exit`, but feels weird

main.go
```diff
@@ -73,6 +73,10 @@ func main() {
 		},
 	)
 	if err := ctx.Run(); err != nil {
+		if errors.Is(err, huh.ErrTimeout) {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(exit.StatusTimeout)
+		}
 		if errors.Is(err, exit.ErrAborted) || errors.Is(err, huh.ErrUserAborted) {
 			os.Exit(exit.StatusAborted)
 		}
```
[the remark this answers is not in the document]
caarlos0: no need to I think 🤔

Issue #676, opened 2024-09-14
gum filter --sort gives unexpected results


hxlnt:
**Describe the bug**
`gum filter --sort` returns an unexpected sort (reverse of as-authored order instead of alphabetical).

**To Reproduce**
1. Run sample filter script after adding `--sort` flag and with the following entries: "Cow," "Cat," "Dog," "Aog," and "Cyg"
2. Filter by typing "c," "a," and/or "g"

**Expected behavior**
1. Results are sorted alphabetically when `--sort` flag is used.

**Screenshots**
![Screenshot from 2024-09-13 23-52-22](https://github.com/user-attachments/assets/c87e561b-42a0-4b15-b998-2be856d9d1f4)

**Desktop (please complete the following information):**
 - OS: Linux Mint

Sorry if I am misunderstanding the purpose of this flag; I assumed `--sort` would sort alphabetically rather than reversing the as-authored order.

piero-vic:
It seems like the `--sort` flag is set to [true by default](https://github.com/charmbracelet/gum/blob/c00c2e416acd374423c5661341cff2e50ce23214/filter/options.go#L37C1-L37C113). So there's no difference between running `gum filter` or `gum filter --sort` in this case.

`gum filter` also depends on [fuzzy](https://github.com/sahilm/fuzzy) for the fuzzy finding. And for what I've read it sorts the results by matching quality not alphabetically.

If you would like to disable sorting you can use the `--no-sort` flag and it will preserve the as-authored order.

 You can also sort the input alphabetically with the `sort` command before piping the output to `gum filter`.

```bash
cat input.txt | sort | gum filter --no-sort 
```

hxlnt:
@piero-vic I did think of that, but setting `fuzzy` to false and `sort` to true doesn't sort the results, either. 

Also, when fuzzy matches are of equal quality and `--sort` is true/unset, the results return in *reverse-authored order.* That seems quite strange.

So, I'm just not sure what `--sort` is supposed to do and if it's working as intended or not. At minimum, some documentation might help folks like myself understand the intended behavior. 

Here is a table to summarize some testing I did. 

| Sort flag | Fuzzy flag | Filter behavior | Expected or unexpected? |
|--------------|---------------|-----------------------|---------------------------------------|
| `--sort` | `--fuzzy` | Matches are in fuzzy order, but when fuzzy match quality is equal**, results are in *reverse-authored* order | Unexpected |
[...]

PR #738, opened 2024-12-03
fix: clarify filter --sort flag

caarlos0:
refs https://github.com/charmbracelet/gum/issues/676


filter/options.go
```diff
@@ -34,6 +34,9 @@ type Options struct {
 	Value                 string        `help:"Initial filter value" default:"" env:"GUM_FILTER_VALUE"`
 	Reverse               bool          `help:"Display from the bottom of the screen" env:"GUM_FILTER_REVERSE"`
 	Fuzzy                 bool          `help:"Enable fuzzy matching" default:"true" env:"GUM_FILTER_FUZZY" negatable:""`
```
meowgorithm, line 36:
```suggestion
	Fuzzy                 bool          `help:"Enable fuzzy matching; otherwise match from start of word" default:"true" env:"GUM_FILTER_FUZZY" negatable:""`
```

PR #742, opened 2024-12-05
refactor: removing huh as a dep

### LINKED ISSUE #741: refactor: remove huh as a dependency
`gum choose`
- add header
- add filter
combine `gum choose` and `gum filter`
- full screen option
- select/unselect all

bash to check what bugs have been fixed in prev versions of huh and gum to test against this new branch

caarlos0:
fixes #741

### DIFF
--- a/confirm/confirm.go
+++ b/confirm/confirm.go
@@ -1,0 +1,121 @@
+// Package confirm provides an interface to ask a user to confirm an action.
+// The user is provided with an interface to choose an affirmative or negative
+// answer, which is then reflected in the exit code for use in scripting.
+//
+// If the user selects the affirmative answer, the program exits with 0. If the
+// user selects the negative answer, the program exits with 1.
+//
+// I.e. confirm if the user wants to delete a file
+//
+// $ gum confirm "Are you sure?" && rm file.txt
+package confirm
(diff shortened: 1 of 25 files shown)


aymanbagabas (approved):
Yass!! 🙌

meowgorithm (approved):
_BEAST_!

confirm/confirm.go
```diff
@@ -0,0 +1,206 @@
...
+	Help        key.Binding
+}
+
+// FullHelp implements help.KeyMap.
+func (k keymap) FullHelp() [][]key.Binding {
+	return [][]key.Binding{
+		{k.Toggle, k.Submit, k.Affirmative, k.Negative},
+		{k.Abort, k.Quit, k.Help},
+	}
+}
```
caarlos0, line 82: i wonder if need full help? didn't add to the other components, I think we either add it in all or none of them, ideally?

wdyt @meowgorithm

meowgorithm: Hmm. I think you're right that it's generally extraneous, though in some cases I think it's helpful. Why don't we hide it by default across the board and eventually aim to get it into all the components that make sense?

PR #745, opened 2024-12-06
Quote tmux session name in code sample

oalders:
A tmux session name could contain a space. Quote the variable to avoid
the session name being split on spaces.

### Changes
- Update tmux session picker example in README to allow for session names which contain spaces


meowgorithm (approved):
Solid bash skills. Well done.

caarlos0:
thank you!

PR #747, opened 2024-12-06
feat: improve handling ctrl+c and timeouts

caarlos0:
- make `ctrl+c` interrupt and exit properly
- handle timeouts with a `context.Context` and exit properly


choose/choose.go
```diff
@@ -134,27 +128,13 @@ type item struct {
 }
 
 func (m model) Init() tea.Cmd {
-	return timeout.Init(m.timeout, nil)
+	return nil
 }
```
ccoVeille, line 134: Do you still need Init then?

confirm/confirm.go
```diff
@@ -110,7 +103,7 @@ type model struct {
 }
 
 func (m model) Init() tea.Cmd {
-	return timeout.Init(m.timeout, m.defaultSelection)
+	return nil
 }
```
ccoVeille, line 107: Do you still need it?

pager/command.go
```diff
@@ -40,17 +49,20 @@ func (o Options) Run() error {
...
+		tea.WithContext(ctx),
+	).Run()
 	if err != nil {
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			return exit.ErrTimeout
+		}
+		if errors.Is(err, tea.ErrInterrupted) {
+			return exit.ErrAborted
+		}
 		return fmt.Errorf("unable to start program: %w", err)
```
ccoVeille, line 65: Shouldn't this logic moved in the root command or main?

It's not logic you are copy pasting it everywhere the same dozen lines

But then it means the context should be created at top level and propagated, and not one created in each command

pager/pager.go
```diff
@@ -28,25 +25,14 @@ type model struct {
...
 	maxWidth            int
-	timeout             time.Duration
-	hasTimeout          bool
-	timedOut            bool
 }
 
 func (m model) Init() tea.Cmd {
-	return timeout.Init(m.timeout, nil)
+	return nil
 }
```
ccoVeille, line 32: Remove?

ccoVeille (commented):
I know it's a draft, but I looked at it

choose/choose.go
caarlos0: yes

confirm/confirm.go
caarlos0: yes - need to be a tea.Model

pager/command.go
caarlos0: yup, just did that :)

pager/pager.go
caarlos0: can't, need to implement tea.Model

ccoVeille (approved):

choose/choose.go
```diff
@@ -133,28 +127,12 @@ type item struct {
...
-func (m model) Init() tea.Cmd {
-	return timeout.Init(m.timeout, nil)
-}
+func (m model) Init() tea.Cmd { return nil }
 
 func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.WindowSizeMsg:
 		return m, nil
-	case timeout.TickTimeoutMsg:
```
aymanbagabas, line 144: Is this a breaking change?

pager/pager.go
```diff
@@ -144,17 +130,14 @@ func (m model) KeyHandler(key tea.KeyMsg) (model, func() tea.Msg) {
...
 
 func (m model) View() string {
-	var timeoutStr string
-	if m.hasTimeout {
-		timeoutStr = timeout.Str(m.timeout) + " "
-	}
-	helpMsg := "\n" + timeoutStr + " ↑/↓: Navigate • q: Quit • /: Search "
+	// TODO: use help bubble here
```
aymanbagabas, line 133:
```suggestion
```

aymanbagabas (approved):
Looks good, few questions

pager/pager.go
caarlos0: its in another PR

choose/choose.go
caarlos0: shouldn't be no

bashbunni:
This works great for the Gum commands, though when working with a script (e.g. `demo.sh`) hitting `ctrl+c` skips the current gum command, but it would be great to be able to quit the whole script with `ctrl+d`. The changes here work as expected :) would be great to hear your thoughts on how we could exit the script though without having to skip all of the remaining gum commands

caarlos0:
@bashbunni it does work, but the script might need the `set -euo pipefail` options, e.g.:

```bash
#!/bin/bash
set -euo pipefail
time=$(gum input --placeholder "Timeout (in minutes)")
gum spin --timeout "${time}m" --show-output -- ping -c 50 google.com
echo "Done"
```

- `-e`: exits on non-zero status
- `-u`: errors on unset variables (not necessary for this example though)
- `-o pipefail`: if a command in a pipe fails, returns its status code

these options are generally good to have on any scripts as they help preventing errors.

choose/choose.go
caarlos0: FWIW `--timeout` isn't even working in latest gum version  :P

PR #748, opened 2024-12-06
fix(pager): use help bubble

caarlos0:
use the help bubble in the pager component

### DIFF
--- a/pager/command.go
+++ b/pager/command.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 
+	"github.com/charmbracelet/bubbles/help"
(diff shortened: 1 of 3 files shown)


aymanbagabas (approved):

bashbunni (approved):
before:
image

after:
image

bashbunni:
didn't look at why lint is failing though 🤔

caarlos0:
> didn't look at why lint is failing though 🤔

was fixed in previous PRs I think

caarlos0:
> before: image
> 
> after: image

yup, only change was to make the help lowercase, so it matches with the other components as well

PR #752, opened 2024-12-09
fix(examples): remove -e from echo

### LINKED ISSUE #751: -e inserted into demo.sh
image

image

reproduce: run `sh demo.sh` on `main` branch. Some pages show `-e` in the output

macOS, kitty, fish shell
edit: tried on ghostty with zsh and still encountered this issue

bashbunni:
Fixes #751 

I don't think `-e` is needed since we use `gum style`? I don't encounter this issue anymore
image

### DIFF
--- a/examples/demo.sh
+++ b/examples/demo.sh
@@ -3,11 +3,11 @@
 gum style --border normal --margin "1" --padding "1 2" --border-foreground 212 "Hello, there! Welcome to $(gum style --foreground 212 'Gum')."
 NAME=$(gum input --placeholder "What is your name?")
 
-echo -e "Well, it is nice to meet you, $(gum style --foreground 212 "$NAME")."
+echo "Well, it is nice to meet you, $(gum style --foreground 212 "$NAME")."
 
 sleep 1; clear
 
-echo -e "Can you tell me a $(gum style --italic --foreground 99 'secret')?\n"
+echo "Can you tell me a $(gum style --italic --foreground 99 'secret')?\n"
 
 gum write --placeholder "I'll keep it to myself, I promise!" > /dev/null # we keep the secret to ourselves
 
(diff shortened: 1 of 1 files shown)


examples/demo.sh
```diff
@@ -3,11 +3,11 @@
 gum style --border normal --margin "1" --padding "1 2" --border-foreground 212 "Hello, there! Welcome to $(gum style --foreground 212 'Gum')."
 NAME=$(gum input --placeholder "What is your name?")
 
-echo -e "Well, it is nice to meet you, $(gum style --foreground 212 "$NAME")."
+echo "Well, it is nice to meet you, $(gum style --foreground 212 "$NAME")."
 
 sleep 1; clear
 
-echo -e "Can you tell me a $(gum style --italic --foreground 99 'secret')?\n"
+echo "Can you tell me a $(gum style --italic --foreground 99 'secret')?\n"
```
caarlos0, line 10: here I believe it is needed because of the trailing `\n` 🤔

examples/demo.sh
```diff
@@ -24,7 +24,7 @@ grep -q "$DISCARD" <<< "$ACTIONS" && gum spin -s monkey --title " Discarding you
 
 sleep 1; clear
 
-GUM=$(echo -e "Cherry\nGrape\nLime\nOrange" | gum filter --placeholder "Favorite flavor?")
+GUM=$(echo "Cherry\nGrape\nLime\nOrange" | gum filter --placeholder "Favorite flavor?")
```
caarlos0, line 27: here also, the `\n`s

meowgorithm:
To Carlos's point, I believe we need the `-e` because of the newlines. Also note that some people would argue that `printf` is more correct.

caarlos0:
also just looked into the pointed issue, it seems to me that shouldn't happen... and in fact I cannot reproduce it either... not sure how this could happen 🤔

bashbunni:
@caarlos0 did you try to repro on Mac or Linux?

caarlos0:
@bashbunni both 🤔

bashbunni:
oh heh my bad, it was because I ran the example with `sh demo.sh` so must not be supported as a flag for `echo` in `sh`

Issue #666, opened 2024-09-02
Gum write limited to 99 lines despite --char-limit 0


fbnlbl-hws:
**Describe the bug**
Gum write limited to 99 lines despite --char-limit 0.
Gum version: v0.14.4 (9c11289)

**To Reproduce**

`yes "HelloWorld" | head -n 110 | gum write --show-line-numbers --char-limit 0`

**Expected behavior**
110 lines of HelloWorld in the open editor.

**Screenshots**
![image](https://github.com/user-attachments/assets/247363a9-3c52-4260-af98-0972f61afe93)

**Desktop (please complete the following information):**
 - OS: Red Hat Enterprise Linux 8.10 (Ootpa)
 - Kernel: Linux 4.18.0-553.5.1.el8_10.x86_64
 - Gum version: gum version v0.14.4 (9c11289)

fbnlbl-hws:
The max line count of 99 is apparently based on the default 2d rune slice in bubbles textarea New() function.
It uses the `defaultMaxHeight` as a line count during initialization, which is set to 99.

[textarea.go#L287](https://github.com/charmbracelet/bubbles/blob/master/textarea/textarea.go#L287)


This seems to be a hard coded line maximum for the gum write tool.
Would be great if it were possible to set this value via flags.

PR #753, opened 2024-12-09
fix(write): max height, max chars

caarlos0:
- it makes no sense to me to have 400 as limit?
- textarea default max height is 99, we should probably set it to 0

closes https://github.com/charmbracelet/gum/issues/666


meowgorithm (approved):
LGTM! Fwiw, in v2 we're removing these baked-in limits in `textarea` itself.

PR #754, opened 2024-12-09
fix(pager): do not strip ansi sequences

### LINKED ISSUE #509: Highlighting doesn't seem to work when combining `glow` with `gum pager`
**Describe the bug**
Highlighting doesn't appear to work when using `gum pager` as the pager for `glow`.

**To Reproduce**
Open file
```sh
PAGER="gum pager" glow README.md --pager
```
and then try to search (`/`)

Possibly related issue. I don't know if this is even supported behavior, but trying to search after running this

```sh
glow README.md | gum pager
```

causes a panic `runtime error: slice bounds out of range [13:10]`

but

```sh
glow README.md | less -r
```

works fine.

**Expected behavior**
Match highlighting style works similar to `less`.

**Screenshots**

Using default `less -r`
```sh
glow -s dark --pager README.md
```
and then `/o|d`, I get the following.

less-matches

Doing the same thing with `gum pager`
```sh
PAGER="gum pager" glow -s dark --pager README.md
```
gum-matches

**Desktop:**
 - OS: Mac OS X Sonoma 14.3.1
 - Terminal: built in Terminal.app
 - Version
   - gum version 0.13.0
   - glow version 1.5.1

caarlos0:
it makes sense to strip everywhere else, but I think the pager is the exception.

closes #509

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -27,7 +27,7 @@ func (o Options) Run() error {
 	)
 
 	if len(o.Options) <= 0 {
-		input, _ := stdin.Read()
+		input, _ := stdin.ReadStrip()
 		if input == "" {
 			return errors.New("no options provided, see `gum choose --help`")
 		}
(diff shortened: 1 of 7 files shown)


bashbunni:
I think this makes sense too. Existing ANSI sequences wouldn't affect usability the same way it does in other components

PR #758, opened 2024-12-10
fix(table): set widths

### LINKED ISSUE #285: Various gum table width bugs
`gum table` has various width bugs

## Width is calculated on header only

```
% gum table <<< "a,b" <<< "1111111111111111111111111,222222222222222222222222"                
 a  b 
 …  … 
```

Since a table has aligned data, gum already needs to read the whole input. Why does gum only use the header text to determine the width of cells by default? It could use the max width of each text column, (perhaps conditionnally to fit the terminal). But the default is not very good.

## `cell.width` still truncates cell text

```
% gum table <<< "a,b" <<< "1111111111111111111111111,222222222222222222222222" --cell.width=20
 a  b 
 …                   …                  
```

why does gum print `…` though there are ~20 empty space characters that could display the `11111…` and `22222…`?

## `cell.width` does not affect header

```
% gum table <<< "a,b" <<< "1111111111111111111111111,222222222222222222222222" --cell.width=20
 a  b 
 …                   …                  
```

resulting in broken alignment between cells and header

## Desktop (please complete the following information):
 - OS: debian
 - Version https://github.com/charmbracelet/gum/releases/tag/v0.9.0

caarlos0:
I'm not sure about this, as it might be slow in big tables...

This will go through all rows, calculate their widths, and set it as the column width if none was provided with `-w`.

```bash
# Without this patch:
echo -e "a,b\naaaaaaaaaaaaaaaaaaaaaaa,bbbbbbbbbbbbbbbb" | gum table
```

```
 a  b
 …  …
```

```bash
# With this patch:
echo -e "a,b\naaaaaaaaaaaaaaaaaaaaaaa,bbbbbbbbbbbbbbbb" | gum table
```

```
 a                        b
 aaaaaaaaaaaaaaaaaaaaaaa  bbbbbbbbbbbbbbbb
```

```bash
# In both:
echo -e "a,b\naaaaaaaaaaaaaaaaaaaaaaa,bbbbbbbbbbbbbbbb" | gum table -w 5,5
```

```
 a      b
 aaaa…  bbbb…
```

closes #285

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -89,6 +89,14 @@ func (o Options) Run() error {
 		if len(row) > len(columns) {
 			return fmt.Errorf("invalid number of columns")
 		}
+		for i, col := range row {
+			if len(o.Widths) == 0 {
+				width := lipgloss.Width(col)
+				if width > columns[i].Width {
+					columns[i].Width = width
+				}
+			}
+		}
 		rows = append(rows, table.Row(row))
 	}
 


meowgorithm (approved):
_Awesome._

PR #761, opened 2024-12-10
fix(stdin): trim space instead of \n

### LINKED ISSUE #682: Unexpected handling of ascii input text
**Describe the bug**
Text emitted by curl when printing headers appear to break line processing. Even after stripping, it's not handled well

**To Reproduce**
This example has embedded terminal handling, and at first I thought this was the problem.
1. `curl -I -i -s http://example.com | gum style --border=normal`

However [stripping](https://superuser.com/a/380778) terminal escapes don't appear to help which is confusing:
1. `curl -I -i -s http://example.com | perl -pe 's/\e\[[0-9;]*m(?:\e\[K)?//g | gum style --border=normal'

**Expected behavior**
Text is formatted correctly.

**Screenshots**
image

 - OS: OSX
 - Version: 14.5
 - Terminal: Alacritty

**Additional context**
Terminal escape stripping is sufficiently complicated I kinda would love to see `gum style --strip-input` as a possible feature!

caarlos0:
we were trimming an ending \n, but it would then break a \r\n sequence, causing misrenders.

this fixes it

closes #682

### DIFF
--- a/internal/stdin/stdin.go
+++ b/internal/stdin/stdin.go
@@ -30,7 +30,7 @@ func Read() (string, error) {
 		}
 	}
 
-	return strings.TrimSuffix(b.String(), "\n"), nil
+	return strings.TrimSpace(b.String()), nil
 }
 
 // ReadStrip reads input from an stdin pipe and strips ansi sequences.


internal/stdin/stdin.go
```diff
@@ -30,7 +30,7 @@ func Read() (string, error) {
 		}
 	}
 
-	return strings.TrimSuffix(b.String(), "\n"), nil
+	return strings.TrimSpace(b.String()), nil
```
ccoVeille, line 33: So you are also trimming on the left.

Please confirm if it intended and OK

aymanbagabas, line 33: Would this trim `\r\n` suffixes?

caarlos0: Yes!

caarlos0: Thought about that, I'm not sure why we were trimming the eof \n... in any case, it's probably fine 🤔

meowgorithm (approved):

Issue #491, opened 2024-02-02
Running tests with gum


dzonekl:
Hi, I am embedding gum into a new modular zsh framework, which also runs tests on zsh functions. 
These functions use `gum choose` and `gum confirm` for example and I would like to be able to automatically 
confirm or choose an item from the list. Is there a way to do that ? 

For example using `yes` command

```zsh
# Should this work ? 
yes | gum confirm 
```

For a choice in a list 

```zsh
colors=(blue yellow) 
echo 'blue' | gum choose $colors
```

hxlnt:
+1.

`gum confirm` and `gum choose` do offer a timeout flag which is helpful for running automated tests. That will essentially just pick the default selection after a certain time has passed, so you'd have to write your tests accordingly.

However, there are some formats that don't have a timeout flag like `gum table`. Not sure how to get these to pass automated tests. Piping in `yes ""` doesn't work, but it would be great if it did!

caarlos0:
added to both confirm and choose

thanks!

PR #772, opened 2024-12-12
feat: yes|gum confirm

caarlos0:
refs #491


aymanbagabas (approved):

bashbunni:
Do we just need `yes` or support for `no` as well? Works great as-is for `yes | gum confirm` (tried various uppercase & lowercase combos of 'YES', all worked great)

caarlos0:
@bashbunni no would be anything that isn't `yes` 🤔

PR #773, opened 2024-12-12
feat(choose): select from stdin

caarlos0:
refs https://github.com/charmbracelet/gum/issues/491

I'm not sure how useful this really is, though...


aymanbagabas (approved):

bashbunni:
Not sure about why we need this either. Maybe we can ask the author of the issue to clarify the value add for this feature?

PR #774, opened 2024-12-12
feat(spin): --show-stdout --show-stderr

### LINKED ISSUE #362: Add an option to ignore stderr when running gum spin
I'm trying to run a program with `gum spin` and redirect `stderr` to `/dev/null` as mention [here](https://github.com/charmbracelet/gum/pull/64#issuecomment-1201497074) but this approach also redirects the output from `gum spin`.

demo

I think an option called `--ignore-error` would be a good idea.

caarlos0:
closes #362

### DIFF
--- a/spin/command.go
+++ b/spin/command.go
@@ -24,7 +24,8 @@ func (o Options) Run() error {
 		title:      o.TitleStyle.ToLipgloss().Render(o.Title),
 		command:    o.Command,
 		align:      o.Align,
-		showOutput: o.ShowOutput && isTTY,
+		showStdout: (o.ShowOutput || o.ShowStdout) && isTTY,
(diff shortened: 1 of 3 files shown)


bashbunni:
Quick Q: while testing this I'm finding spin isn't animating nor printing the command's output. Maybe user error? Getting this regardless of whether I use a `--show-x` flag or not. It just prints the first render of the spinner

image

caarlos0:
I think it needs to be 

```sh
gum spin --show-stdout --title "Running command..." -- sleep 10 && echo hello
```

for me the command you provided exits 1 🤔

bashbunni:
Ah I get it now! Thank you :D Looks great

PR #775, opened 2024-12-13
feat(version): adds command to check current gum version

### LINKED ISSUE #352: Allow specifying the min. version of `gum` required
The nature of CLI tooling means that the version of `gum` being used is very important - older versions cannot know about the new behaviours of newer versions, and attempting to use an older version of `gum` can result in an error that presents itself as the user being in the wrong (i.e. "--flag is not valid").

Provide a command that allows specifying the version of `gum` required, that people could stick at the start of their scripts and would print a message explaining that the version of `gum` that would be used is not expected to be compatible, and potentially link / explain how to update.

i.e.

```
#!/bin/bash

gum version '0.12.0' # or '>= 0.12.0'

gum style --border normal --margin "1" --padding "1 2" --border-foreground 212 "Hello, there! Welcome to $(gum style --foreground 212 'Gum')."
NAME=$(gum input --placeholder "What is your name?")

echo -e "Well, it is nice to meet you, $(gum style --foreground 212 "$NAME")."
```

This could result in something like:

```
> gum/examples/demo.sh
This script requires gum 0.12.0, but you're using v0.8.0 - it would be best if you update to ensure you don't have any problems!
```

I think ideally it would be nice to support at least a basic constraint rather than requiring just a single specific version, but I think that could be supported later on.

I could attempt to check the version of `gum` using built-ins, but I expect that will be tedious, potentially brittle, and just annoyingly large if we want to have a nice error message.

**Additional context**
While this kind of command itself wouldn't be present in earlier versions of `gum`, the earlier the base command is added the easier it should be in the long-run to expand its support (i.e. if you just started with supporting static versions, the error message could mention "if you're trying to do constraints, you probably want a newer version of `gum`")

caarlos0:
closes #352

### DIFF
--- a/version/command.go
+++ b/version/command.go
@@ -16,7 +16,7 @@ func (o Options) Run(ctx *kong.Context) error {
 	current := ctx.Model.Vars()["versionNumber"]
 	v, err := semver.NewVersion(current)
 	if err != nil {
-		return fmt.Errorf("could parse version %s: %w", current, err)
+		return fmt.Errorf("could not parse version %s: %w", current, err)
 	}
 	if !c.Check(v) {
 		return fmt.Errorf("gum version %q is not within given range %q", current, o.Constraint)
--- a/version/command.go
+++ b/version/command.go
@@ -1,0 +1,25 @@
+package version
+
+import (
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/alecthomas/kong"
+)
+
+// Run check that a given version matches a semantic version constraint.
+func (o Options) Run(ctx *kong.Context) error {
+	c, err := semver.NewConstraint(o.Constraint)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


version/command.go
```diff
@@ -0,0 +1,25 @@
...
+// Run check that a given version matches a semantic version constraint.
+func (o Options) Run(ctx *kong.Context) error {
+	c, err := semver.NewConstraint(o.Constraint)
+	if err != nil {
+		return fmt.Errorf("could not parse range %s: %w", o.Constraint, err)
+	}
+	current := ctx.Model.Vars()["versionNumber"]
+	v, err := semver.NewVersion(current)
+	if err != nil {
+		return fmt.Errorf("could parse version %s: %w", current, err)
```
G-Rath, line 19:
```suggestion
		return fmt.Errorf("could not parse version %s: %w", current, err)
```

PR #779, opened 2024-12-13
feat(choose,filter): --input-delimiter --output-delimiter

### LINKED ISSUE #274: Ability Set to set a input delimiter on `gum filter` and a output delimiter when using `--no-limit`
As it stands you must use new lines `\n` as your delimiter in the input list for `gum filter`. Also when you use `--no-limit` the selections are concatenated without a delimiter.

It would be nice to have the ability to set a delimiter for both.

You could have something like this
```
platform=$(echo "js,android,ios" | gum filter --delimiter=",")

# and

platforms=$(echo "js|android|ios" | gum filter --delimiter="|" --no-limit --output-delimiter=",")
```

caarlos0:
allows to change how content from stdin is used, and how results are printed.

one could get around it piping into and from `tr`, but results aren't quite right, especially when `tr '\n' ','` for example, as it'll add an extra `,` in the end of the string.

This makes things a bit cleaner, hopefully.

closes #274

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -13,9 +13,8 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/gum/internal/stdin"
 	"github.com/charmbracelet/gum/internal/timeout"
+	"github.com/charmbracelet/gum/internal/tty"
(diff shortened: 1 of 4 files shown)


meowgorithm:
This is _so_ nice.

PR #739, opened 2024-12-04
fix: strip ansi sequences from stdin

caarlos0:
most programs should not give output with ansi sequences to a pipe, but as some of them do, this will strip them out.

closes #188

Issue #188, opened 2022-10-09
Sanitize input


muesli:
Input containing ANSI sequences is currently not sanitized before display.

**To Reproduce**
1. Provide ANSI sequences as input to `gum filter`
2. Watch them affect rendering

**Expected behavior**
Input needs to be sanitized before displaying.

maaslalani:
I remember thinking about this a while ago. I decided not to sanitize the input as most shell programs sanitize the input if they are being passed as a pipe output to another program.

```
bat README.md | gum filter
```

So, if ANSI sequences are part of the input perhaps they are intentional?

Curious what you think about this, happy to revisit if you think we should sanitize still though.

muesli:
I think we should sanitize per default, otherwise it can lead to weird results (cursor repositioning in the middle of a render etc.), but maybe have an `--ansi` flag to support pre-styled content.

Issue #689, opened 2024-10-17
color codes break gum filter match foreground


DarioDarko:
**Describe the bug**
using escape codes to color text breaks gum filter. typing 'm' 'b' or '[' as first character messes with the color output of match.foreground. disabling match highlighting with `--match.foreground=""` gets rid of the issue, but highlighting would be nice

this also affects other highlighting styles, i think all of them

i could not reproduce the problem in xterm for some reason, alacritty seems to handle color escape codes differently. not sure about other terminal emulators

**To Reproduce**
``
red=$(tput setaf 1)
green=$(tput setaf 2)
reset=$(tput sgr0)

{
	echo "normal text"
	echo "${red}red text${reset}"
	echo "${green}green text${reset}"
} | gum filter
```

**Expected behavior**
highlight colored text without breaking the escape code

**Screenshots**
![image](https://github.com/user-attachments/assets/956e3058-653e-4f0f-b9bf-b56a25c49665)

**Desktop (please complete the following information):**
 - OS: EndeavourOS
 - Terminal: Alacritty
 - Version: gum version 0.14.5 (1917023)

caarlos0:
fixed on main! thanks!

PR #782, opened 2024-12-16
feat: add --[no-]strip-ansi-codes option

srithon:
refs: #739, https://github.com/charmbracelet/gum/issues/188#issuecomment-1272654231

I chose to hide the conditional behind an interface to simplify the call sites and duplication of flag processing across commands. If we add further stdin-related options in the future, this would become more useful, as more logic would be abstracted away. However, I can see the argument against this approach; this pattern isn't used in the rest of the codebase (from what I saw), and adding it for such a small feature might not be worth the added complexity. In that case, it would be pretty easy to get rid of the interface and just inline the conditional in every command.

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -26,7 +26,7 @@ func (o Options) Run() error {
 	)
 
 	if len(o.Options) <= 0 {
-		input, _ := stdin.ReadStrip()
+		input, _ := stdin.ReadWithOptions(&o)
 		if input == "" {
 			return errors.New("no options provided, see `gum choose --help`")
 		}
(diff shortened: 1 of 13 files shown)


srithon:
I realized that the [fix](https://github.com/charmbracelet/gum/issues/689#issuecomment-2529285188) for #689 was #739, and so it'll take more work than just re-enabling ANSI codes to make colored options work correctly.

caarlos0:
did a slightly different implementation in #784

caarlos0:
merged https://github.com/charmbracelet/gum/pull/784

caarlos0:
thanks for the PR btw :)

PR #784, opened 2024-12-17
feat: --no-strip-ansi

caarlos0:
I think this would be mostly useful for choose and filter, but anyway, this allows to disable the ANSI stripping when reading from STDIN.

alternative implementation of #782

### DIFF
--- a/choose/command.go
+++ b/choose/command.go
@@ -27,7 +27,7 @@ func (o Options) Run() error {
 	)
 
 	if len(o.Options) <= 0 {
-		input, _ := stdin.ReadStrip()
+		input, _ := stdin.Read(stdin.StripANSI(o.StripANSI))
 		if input == "" {
 			return errors.New("no options provided, see `gum choose --help`")
 		}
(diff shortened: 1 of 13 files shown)


srithon:
Hey @caarlos0, I mentioned this briefly on #782, but I think fully implementing this feature will require a bit more work; I went into more detail in issue #785.

caarlos0:
for the specific cases of the styling, yes, it might need more work, currently i would classify it as 'least effort' - if you strip ansi, it might work (if its just color changes for example), if its more complex things, it probably won't.

can you share some specific case in which you found problems?

srithon:
> for the specific cases of the styling, yes, it might need more work, currently i would classify it as 'least effort' - if you strip ansi, it might work (if its just color changes for example), if its more complex things, it probably won't.
> 
> can you share some specific case in which you found problems?

Sorry for the late reply, but I gave an example in issue #785. I could try to help resolve the issue, but unfortunately I'm not sure where to start with this one.

Issue #785, opened 2024-12-18
--no-strip-ansi prints ANSI codes and garbles output


srithon:
This flag was introduced in PR #784, after a commit in PR #739 unconditionally stripped ANSI codes from stdin. PR #739 was made in response to issue #689; in order to prevent the display from being garbled by incorrect ANSI code handling, we opted to simply strip ANSI codes from the input entirely. The implementation in PR #784 (and my own draft implementation in PR #782) simply undo that stripping behavior without addressing the underlying issues from #689, resulting in strange outputs in certain cases. Below is a minimal example which results in garbled output on my machine in Kitty, Alacritty and st.

```sh
#!/bin/sh

green=$(tput setaf 2)
reset=$(tput sgr0)

{
  echo "Option 1"
  echo "Option 2"
  echo "${green}Option 3${reset}"
} | gum filter --no-strip-ansi
```

Here are recordings for each tested terminal. Note that Kitty was a lot worse than the others; interestingly, when I tried using `asciinema` to record Kitty's output, the recording showed the same `2m` result as the other terminals; to show what it actually looked like on-screen I had to use a screen recording.

1.  [Kitty](https://github.com/user-attachments/assets/1113aef1-95a2-4e0a-992e-71544fbad190)
2.  [Alacritty](https://asciinema.org/a/Za0HsgCy7tleuim9PVLbgUo3e)
3.  [st](https://asciinema.org/a/2V2ZGmVuUq4QftovkQJcGiNvd)

caarlos0:
oh, I see, lemme fix that!

caarlos0:
to resolve this properly I think we need 2 things:

1. matching et al against the ansi-stripped string (just implemented this in a local branch)
2. when printing the matched options, we'll need to deal with ansi sequences when highlighting as well - gonna try to implement this into x/ansi

caarlos0:
is this what you expect would happen?

![CleanShot 2025-01-06 at 16 17 57@2x](https://github.com/user-attachments/assets/ed834cf1-f02d-4c05-9ce8-6495e86dbd09)

srithon:
> is this what you expect would happen?
> 
> 
> 
> ![CleanShot 2025-01-06 at 16 17 57@2x](https://github.com/user-attachments/assets/ed834cf1-f02d-4c05-9ce8-6495e86dbd09)
> 
> 

This looks perfect, thank you! I'll test it out and let you know if I find anything else.

joshmedeski:
[...]

PR #789, opened 2025-01-06
fix(filter): properly handle options with ansi styles

caarlos0:
Fixes https://github.com/charmbracelet/gum/issues/785

- it will fuzzy match against the stripped string 
- it will then render the styles string, taking ansi escape codes into account
- when a char matches, it'll use the matched style instead of whatever style the item had previously
- other chars will have their previous style though
- refactored the matches styling to do it in ranges instead of per char basis, as we now need to `ansi.Truncate`/`ansi.TruncateLeft` to avoid breaking styles

some screenshots:

CleanShot 2025-01-06 at 17 17 09@2x
CleanShot 2025-01-06 at 17 17 30@2x
CleanShot 2025-01-06 at 17 17 38@2x
CleanShot 2025-01-06 at 17 17 42@2x

### DIFF
--- a/filter/command.go
+++ b/filter/command.go
@@ -14,6 +14,7 @@ import (
 	"github.com/charmbracelet/gum/internal/stdin"
 	"github.com/charmbracelet/gum/internal/timeout"
 	"github.com/charmbracelet/gum/internal/tty"
+	"github.com/charmbracelet/x/ansi"
 	"github.com/sahilm/fuzzy"
 )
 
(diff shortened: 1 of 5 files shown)


meowgorithm:
Awesome. I'm having a hard time reproducing this one. What am I doing wrong?

```bash
printf "\x1b[38;5;212;48;5;63mOption 1,Option 2,Option3\x1b[m" | go run . filter --input-delimiter="," --height="10"
```

caarlos0:
missing `--no-strip-ansi`

see https://github.com/charmbracelet/gum/issues/785 for a reproducible

PR #806, opened 2025-01-16
feat(table): Add interactive flag for #800

### LINKED ISSUE #800: gum table doesn't allow selecting on stdout redirect
**Describe the bug**
When using `gum table` to store the result in a variable or redirecting it to a file, it just prints the table.

**To Reproduce**
1. `echo -e "a,b,c\n1,2,3\n4,5,6" | gum table`
2. `echo -e "a,b,c\n1,2,3\n4,5,6" | gum table | cat`

**Expected behavior**
I would expect both version to behave identical, displaying the table, let me choose a line and printing the selected line. The behaviour to print the selected line only in interactive mode is kinda useless.

 - OS: macOS and Linux tested
 - Version gum version 0.15.0

MTHLily:
Fixes #800 

### Changes
- Add `--interactive/-i` flag to restore previous functionality to select a table row then pipe it into another command

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -110,7 +110,7 @@ func (o Options) Run() error {
 		rows = append(rows, table.Row(data[row]))
 	}
 
-	if !o.Interactive && (o.Print || !term.IsTerminal(os.Stdout.Fd())) {
+	if !o.Interactive && (o.Print || !term.IsTerminal(os.Stdin.Fd())) {
 		table := ltable.New().
 			Headers(columnNames...).
 			Rows(data...).
--- a/table/command.go
+++ b/table/command.go
@@ -110,7 +110,7 @@ func (o Options) Run() error {
 		rows = append(rows, table.Row(data[row]))
 	}
 
-	if o.Print || !term.IsTerminal(os.Stdout.Fd()) {
+	if !o.Interactive && (o.Print || !term.IsTerminal(os.Stdout.Fd())) {
 		table := ltable.New().
 			Headers(columnNames...).
 			Rows(data...).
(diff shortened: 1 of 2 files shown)


table/command.go
```diff
@@ -110,7 +110,7 @@ func (o Options) Run() error {
 		rows = append(rows, table.Row(data[row]))
 	}
 
-	if o.Print || !term.IsTerminal(os.Stdout.Fd()) {
+	if !o.Interactive && (o.Print || !term.IsTerminal(os.Stdout.Fd())) {
```
caarlos0, line 113: I think maybe we should check if stdin is a terminal instead 🤔

MTHLily: Implemented your suggestions, @caarlos0. However, the behavior is inconsistent for the following cases:

```sh
results="$(gum table --height 10 -f test.csv)" # Results in an interactive table
results="$(gum table --height 10 <test.csv)" # Results in a printed table
results="$(cat test.csv | gum table --height 10)" # Results in a printed table
```

MTHLily: Reflecting on the way `gum table` is meant to be used and described in the readme, i.e. as a row selector for tabular data, I'm thinking that instead of adding the forced interactive flag, we ought to revert the original commit from #762. I don't think it's expected behavior for `gum table` to print a table instead of letting us choose a row.

Cards on the table, I was also using `gum table` to select row data and piping it to another program for my own script. So, I am a bit biased in this regard.

caarlos0: yes, in retrospect, I think that it makes sense. will do it

caarlos0: https://github.com/charmbracelet/gum/pull/811

aymanbagabas (approved):

bashbunni:
Hey, looks like this was closed in #811. Thanks so much for the contribution :) If you have any questions or concerns about this PR, please let us know and we can reopen.

PR #810, opened 2025-01-19
chore(nix): update src hash and nix version

ctn-malone:
### Changes
- changed src hash to make main branch build through nix 
- change version to 0.15.0
- update nix lock file to have go >= 1.23 instead of 1.22.2


caarlos0:
thanks!

Issue #803, opened 2025-01-15
Gum spinner only shows the title and the dots are not visible


thesobercoder:
**Describe the bug**
Gum spinner only shows the title and the dots are not visible

**To Reproduce**
1. Make sure that AWS CLI installed and logged in.
2. Make sure you have `jq` intalled.
3. Run the following command.
   ```sh
   gum choose --header "Choose region:" $(gum spin --spinner dot --title "Fetching regions" --show-output -- aws ec2 describe-regions --region "us-east-1" --output json | jq -r ".Regions[].RegionName")
   ```

**Expected behavior**
The spinner should show correctly beside the title.

**Screenshots**
![CleanShot 2025-01-15 at 23 34 26](https://github.com/user-attachments/assets/4a455e0b-2450-430d-92f5-689880a5506b)

**Desktop (please complete the following information):**
 - OS: macOS 15.2
 - Terminal: Ghostty 1.0.1
 - jq: jq-1.6-159-apple-gcff5336-dirty
 - Gum: 0.15.0

diegodorado:
Got a similar experience... title shows, but dots doesn't

any kind of piping breaks:

```
A=$(gum spin -- curl -s https://api.github.com/users/hadley/orgs)
```

or

```
gum spin -- curl -s https://api.github.com/users/hadley/orgs | jq
```

caarlos0:
should be fixed in #813

diegodorado:
Nice! it works now on gum version 0.15.1
Thanks!

caarlos0:
Thanks for the report! :)

PR #812, opened 2025-01-21
Allow to show spinner on non tty output devices

ctn-malone:
Adds a `--show-spinner` option to work around the fact that by default, spinner won't be shown when output device is not a TTY (ie: redirected or piped) on version `0.15.0` (introduced in https://github.com/charmbracelet/gum/commit/cf2da6406cbb2827e460cfd2cf9a2c20ed3affeb)

**Examples**

Screenshot_20250121_135735

Screenshot_20250121_135836

While flag name may be a bit confusing, I think such option may be useful and would fix some issues such as https://github.com/charmbracelet/gum/issues/803

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,8 @@ automatically stop after the given command exits.
 
 To view or pipe the command's output, use the `--show-output` flag.
 
+By default, spinner will not be printed when output is piped or redirected. Use the `--show-spinner` to make it visible in such case.
+
 ```bash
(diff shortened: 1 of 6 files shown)


caarlos0:
hmm, I think the problem there is that we should have checked for stderr instead of stdout, let me give it a better look

caarlos0:
can you try #813?

ccoVeille (approved):

ctn-malone:
> can you try #813?

Yep, much better indeed :). Closing in favor of #813

PR #813, opened 2025-01-21
fix: spin when not a tty

caarlos0:
fixes #812
refs #763

### DIFF
--- a/spin/command.go
+++ b/spin/command.go
@@ -14,7 +14,7 @@ import (
 // Run provides a shell script interface for the spinner bubble.
 // https://github.com/charmbracelet/bubbles/spinner
 func (o Options) Run() error {
-	isOutTTTY := term.IsTerminal(os.Stdout.Fd())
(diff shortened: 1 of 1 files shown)


spin/command.go
```diff
@@ -14,7 +14,8 @@ import (
 // Run provides a shell script interface for the spinner bubble.
 // https://github.com/charmbracelet/bubbles/spinner
 func (o Options) Run() error {
-	isTTY := term.IsTerminal(os.Stdout.Fd())
+	isOutTTTY := term.IsTerminal(os.Stdout.Fd())
```
bashbunni, line 17: should this be `isOutTTY` or is the `TTTY` intentional?

bashbunni (approved):
Looks good. Does this affect tests?

spin/command.go
bashbunni: Looks great :)

demo

PR #791, opened 2025-01-07
horizontal scrolls and other improvements

caarlos0:
pager:
- search input width 
- horizontal scroll

filter:
- horizontal scroll

PR #827, opened 2025-01-28
fix(pager): memory/cpu usage when using soft-wrap

### LINKED ISSUE #823: memory leak in pager
**Describe the bug**
`gum pager` no longer functions correctly as $PAGER, and seems to have a memory leak

**To Reproduce**
1. in 0.15.1: in zshrc, have `export PAGER="gum pager"`
2. open a command that would automatically use your pager, such as `man tail`
3. gum pager opens, but no content is displayed, and the entire page is a small pink square with the controls displayed.
4. gum pager uses 100% of 1 CPU, and fills the entire available ram in a few minutes. the server becomes understandably unresponsive and needs to be restarted.

**Expected behavior**
in 0.14-, this exact configuration worked as expected. `man tail` would correctly display inside gum pager without issues.

**Screenshots**

Image

 - OS: Almalinux 8
 - Version 0.15.1

caarlos0:
easily reproducible, especially with something like

```sh
PAGER="gum pager" man ls
```

Problems here were:
- reflow's truncate is not very efficient
- useless calls to truncate
- bad strings.Replace calls (not necessary, and wouldn't work if text has ansi sequences)

Fixes #823 

PS: this is better handled in #791, but that one is a bigger change (and depends on a /bubbles change as well), so I think we can probably do this instead for now and push a patch until #791 is done and merged.

---

#### before:

CleanShot 2025-01-28 at 11 24 21@2x

#### fixed on regular usage, search still broken

image

#### full fix

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/charmbracelet/x/ansi v0.7.0
 	github.com/charmbracelet/x/editor v0.1.0
 	github.com/charmbracelet/x/term v0.2.1
-	github.com/muesli/reflow v0.3.0
 	github.com/muesli/roff v0.1.0
 	github.com/muesli/termenv v0.15.3-0.20241211131612-0d230cb6eb15
 	github.com/rivo/uniseg v0.4.7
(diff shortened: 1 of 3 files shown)


caarlos0:
also worth noting: memory usage, before x after 

CleanShot 2025-01-28 at 11 29 18@2x


CleanShot 2025-01-28 at 11 32 01@2x



PS: first screenshot is just opening gum pager, second is navigating, otherwise there's actually no allocations ;)

meowgorithm (approved):

aymanbagabas (approved):

PR #828, opened 2025-01-28
fix: GUM_DEBUG and pprof

caarlos0:
Used this in #827 and found it quite useful.

Should we ship this to help both we and our users to see mem/cpu usage etc?

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"runtime/debug"
 
-	_ "net/http/pprof"
+	_ "net/http/pprof" //nolint:gosec
 
 	"github.com/alecthomas/kong"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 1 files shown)


aymanbagabas (commented):
This is great! I would still hide it behind a build flag just to avoid adding more dependencies.

caarlos0:
> This is great! I would still hide it behind a build flag just to avoid adding more dependencies.

could do that, but then users still can't debug it without downloading another binary

not sure if this should be a goal or not, but was initially thinking it could be 🤔

aymanbagabas:
> could do that, but then users still can't debug it without downloading another binary
> 
> not sure if this should be a goal or not, but was initially thinking it could be 🤔

Right, I'm just not sure that extra 5MB is worth it :/

```
.rwxr-xr-x@  17M ayman 28 Jan 11:18 gum-main
.rwxr-xr-x@  22M ayman 28 Jan 11:18 gum-pprof
```

caarlos0:
true, probably isn't 

we can also close this and use it only when we need it (so far one time heh)

Issue #196, opened 2022-10-12
Show that more elements are available on scroll


simjnd:
**Is your feature request related to a problem? Please describe.**
Use `gum table < test.csv --height 3` when `test.csv` has 10 rows.  It will display a table that has 3 rows and no way of knowing if you've reached the bottom of the table or if there is more available by "scrolling".

**Describe the solution you'd like**
When there are more elements available, show an indication (e.g. add a row with dark gray `...` showing content has been clipped from the viewport)

**Describe alternatives you've considered**
N/A

**Additional context**
<img width="338" alt="image" src="https://user-images.githubusercontent.com/43272781/195425937-cf6d41aa-ae69-4370-a7c8-0ddce8da67f2.png">

*How many items are there?*

<img width="336" alt="image" src="https://user-images.githubusercontent.com/43272781/195425965-452d936a-529e-47c9-a4ae-8a53582bbecb.png">

*Turns out there are a lot!*

olivertzeng:
I like the idea!

PR #839, opened 2025-02-04
feat: show indicator on help keybindings (opt-in)

raphamorim:
Fixes https://github.com/charmbracelet/gum/issues/196

This allow table display to be configurable and show item quantity on helper. It can be disabled by `hideIndicator`.

## Before

Screenshot 2025-02-03 at 15 05 35

## After

Screenshot 2025-02-04 at 11 39 49

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -144,6 +144,7 @@ func (o Options) Run() error {
 	m := model{
 		table:    table,
 		showHelp: o.ShowHelp,
+		hideIndicator: o.HideIndicator,
 		help:     help.New(),
 		keymap:   defaultKeymap(),
 	}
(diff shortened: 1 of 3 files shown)


aymanbagabas (approved):

FilipeBento:
**_Excellent_** work ❤️! Why not extend it for more cases instead of just `table`?

(replace seq with your use-case)
`seq 100 | gum filter`

`seq 100 | gum choose`

I'd still like to know how many rows I have available and which row I am in. I think this would be useful for more use-cases, like [fzf](https://github.com/junegunn/fzf) does.

Love you guys :)

caarlos0:
yeah that probably makes sense! 

created a discussion for it: https://github.com/charmbracelet/gum/discussions/869

PR #850, opened 2025-02-10
fix(spin): preserve color output when `--show-output` is given

### LINKED ISSUE #305: Preserve output formatting with `spin --show-output`
It looks like `gum spin --show-output` strips stdout of all escape codes.

It'd be great if output would be 1:1 as if run without `gum spin`; colors, newlines, the works.

andreynering:
Fixes #305

Running the command inside a PTY (using [this package](https://github.com/creack/pty)) to ensure colors and general formatting are preserved.


spin/pty.go
```diff
@@ -0,0 +1,30 @@
+package spin
+
+import (
+	"os"
+
+	"github.com/charmbracelet/x/term"
+	ptytty "github.com/creack/pty"
```
aymanbagabas, line 7: Fyi, this wouldn't work on Windows :/ We should use our `x/xpty` instead for a cross-platform solution

raphamorim: yea, i just saw it going to say the same, is `x/pty` conpty?

aymanbagabas: We have `x/xpty` which uses `creack/pty` and our `x/conpty` to provide a cross-platform API

raphamorim: got it! thanks @aymanbagabas

raphamorim (approved):
code lgtm, I was just gonna ask about windows but pinged @andreynering instead

spin/spin.go
```diff
@@ -65,15 +71,48 @@ func commandStart(command []string) tea.Cmd {
...
+				executing.Stderr = errUnixPty.Slave()
+			}
+
+			go io.Copy(io.MultiWriter(&bothbuf, &outbuf), stdoutPty) //nolint:errcheck
+			go io.Copy(io.MultiWriter(&bothbuf, &errbuf), stderrPty) //nolint:errcheck
+
+			if err = stdoutPty.Start(executing); err != nil {
+				return errorMsg(err)
+			}
+			_ = xpty.WaitProcess(context.Background(), executing)
```
aymanbagabas, line 109: Just an FYI, we don't need to use `xpty.WaitProcess` here, we can call `executing.Wait()` directly. That's what it does under the hood.

spin/spin.go
```diff
@@ -65,15 +71,48 @@ func commandStart(command []string) tea.Cmd {
 		}
 
 		executing = exec.Command(command[0], args...) //nolint:gosec
-		if term.IsTerminal(os.Stdout.Fd()) {
+		executing.Stdin = os.Stdin
+
+		isTerminal := term.IsTerminal(os.Stdout.Fd())
+
+		//nolint:nestif
+		if isTerminal && runtime.GOOS == "windows" {
```
aymanbagabas, line 79: Probably a TODO comment explaining what's going on

aymanbagabas (approved):
Few nitpicks, but LGTM otherwise!

spin/spin.go
```diff
@@ -65,15 +71,48 @@ func commandStart(command []string) tea.Cmd {
...
 			executing.Stdout = io.MultiWriter(&bothbuf, &outbuf)
 			executing.Stderr = io.MultiWriter(&bothbuf, &errbuf)
+			_ = executing.Run()
```
andreynering, line 82: After some headache, me and @aymanbagabas decided to keep the old behavior on Windows for now.

(Git Bash was freezing for some reason, although the other terminals worked correctly).

We can reconsider in the future if we find a reliable way to make it work on Windows.

spin/spin.go
andreynering: @aymanbagabas I decided to keep as is so the code is more prepared to work on Windows if we decide to try again in the future. WDYT?

spin/spin.go
andreynering: Done

Issue #876, opened 2025-03-18
The countdown timer in `gum confirm --timeout` is not being displayed.


agilesteel:
**Describe the bug**
`gum confirm --timeout` has been broken for many versions. It was recently fixed in https://github.com/charmbracelet/gum/pull/697 so it works in the sense that `confirm` does time out but the countdown timer is not being displayed.

**To Reproduce**
```bash
gum confirm --timeout=10s "All good?" && echo 'yes' || echo 'no'
```

**Expected behavior**
Many versions ago a countdown timer was displayed.

**Screenshots**

![Image](https://github.com/user-attachments/assets/cc5c7e8d-ccbf-4e77-bbae-eda48a3fa0b2)

**Desktop (please complete the following information):**
 - OS: Windows (WSL) Ubuntu 20.04 and M2 Mac as well
 - Version `0.16.0`

agilesteel:
Anyy updates on this?

huyz:
I would love to see this fixed. A visible countdown is the first and, so far, only reason for me to start using `gum`; otherwise, we already have prompts with timeouts in bash

PR #888, opened 2025-04-11
Detect timeout error and apply default option

mikybars:
This is a quick fix to prove the point made here https://github.com/charmbracelet/gum/issues/876#issuecomment-2792207961 about the timeout condition not being handled properly at the moment


andreynering (approved):
Works perfectly. Thank you!

PR #914, opened 2025-06-09
feat: allow setting tea options 

gustavosbarreto:
This adds support for changing some tea options (`WithFPS` and `WithAltScreen`) via options for all tea-based gum commands.

## Motivation

I'm working on a TUI that spawns many gum instances in parallel, potentially hundreds, depending on the number of users. In this context, reducing FPS is essential because gum/tea default to 60fps, but we found 20fps gives a good balance between responsiveness and CPU usage.

We also needed to force alternate screen mode in some commands that don’t use it by default to improve integration in full window environments.

## Implementation

* adds `gumtea.Options`, a struct that supports configuring `tea.ProgramOptions`
* wraps all calls to `tea.NewProgram()` in a new `gumtea.NewProgram()` helper

Each tea-based command now includes a `Tea` field in its options struct

Note that non-tea commands like `join` remain unaffected and the default behavior is preserved since all new settings are opt-in.

## Justification

This enables fine-grained control over tea runtime behavior in scripts, terminals, and complex TUI environments especially useful when gum is embedded into other tools.


gustavosbarreto:
@caarlos0,

> As a small note, we try to start things as discussions, so we can think about things from a cleaner slate.

Just to confirm, is there a contribution guide I should follow? I’d like to avoid submitting another PR that doesn’t align with the project’s expectations

> Finally, if you don't mind me asking, what's the use case of that many gum instances in parallel? Without more info, it sounds like this could probably be its own app, but anyway, I'm curious :)

I'm working on a sort of terminal multiplexer... that's about all I can say for now :P

caarlos0:
yes I would keep the env only only approach... we can always move it later if it makes sense.

caarlos0 (approved):

PR #916, opened 2025-06-09
feat: add --initial option for cursor positioning and fix filter --selected bug

### LINKED ISSUE #337: Filter/Choose: Set starting position of indicator by value
When using `gum choose` or `gum filter`, it might be that one of the available options will be most commonly chosen,
or makes the most sense based on the current environment, current working directory etc.

I'd like the option to set the intial position of the indicator to a specific value.
This would allow a fast and convenient choice of the most common options, by just hitting Enter repeatedly.
An additional flag like f.e. `--indicator-value="my_value"`. There is probably a better name.

For `filter` I could set the `--value` flag, `choose` has `--selected`.
The problem with this is, that it
1. hides all the other options, and/or
2. forces me to first delete the current filter/selection, in case I *do* want to pick another option after all.

Setting the indicator would have the best of both worlds:
1. Quick selection of the default option.
2. If I don't want the default option, I can immediately start typing to filter the results, without having to clear the filter first.

**Additional context**
It might make sense to also allow setting the starting position by index, instead of the value.
Something like `--indicator-index=3`? For my personal use case this wouldn't be very useful though.

Adding an additional flag also allows combining these flags together for even finer control.
Using `--selected` and `--indicator-value`  together allows listing all the "default" options first pre-selected, and then starting the indicator *after* all the default options, so one doesn't have to move through all the default options first.

Using `--value` and `--indicator-value` together allows pre-filtering to reduce a large amount of options, and then having the indicator set on the most probable choice of those pre-filtered options.
*Problem: What if `--value` filters out the option, that was supposed to be used by `--indicator-value`?*

hanoii:
- Add --initial option to both choose and filter commands for setting initial cursor position
- Fix critical bug in filter command where --selected with limit=1 would lock selection
- When limit=1, --selected now behaves like cursor positioning (same as choose command)
- Add comprehensive tests for initial positioning logic
- Support environment variables GUM_CHOOSE_INITIAL and GUM_FILTER_INITIAL

Resolves #337

Changes:
- choose: Add --initial option for cursor positioning (useful with --no-limit)
- filter: Add --initial option and fix --selected behavior with limit=1
- Both commands now have consistent behavior for --selected when limit=1
- Tests added to ensure correct cursor positioning logic

### DIFF
--- a/choose/options.go
+++ b/choose/options.go
@@ -21,6 +21,7 @@ type Options struct {
 	SelectedPrefix   string        `help:"Prefix to show on selected items (hidden if limit is 1)" default:"✓ " env:"GUM_CHOOSE_SELECTED_PREFIX"`
 	UnselectedPrefix string        `help:"Prefix to show on unselected items (hidden if limit is 1)" default:"• " env:"GUM_CHOOSE_UNSELECTED_PREFIX"`
 	Selected         []string      `help:"Options that should start as selected (selects all if given *)" default:"" env:"GUM_CHOOSE_SELECTED"`
+	Initial          string        `help:"Option that should start with the cursor positioned on it" default:"" env:"GUM_CHOOSE_INITIAL"`
 	SelectIfOne      bool          `help:"Select the given option if there is only one" group:"Selection"`
 	InputDelimiter   string        `help:"Option delimiter when reading from STDIN" default:"\n" env:"GUM_CHOOSE_INPUT_DELIMITER"`
 	OutputDelimiter  string        `help:"Option delimiter when writing to STDOUT" default:"\n" env:"GUM_CHOOSE_OUTPUT_DELIMITER"`
--- a/filter/command.go
+++ b/filter/command.go
@@ -132,11 +132,22 @@ func (o Options) Run() error {
 				continue
 			}
 			if o.Limit == 1 {
+				// When the user can choose only one option don't select the option but
+				// start with the cursor hovering over it.
 				m.cursor = i
-				m.selected[option.Str] = struct{}{}
 			} else {
 				currentSelected++
 				m.selected[option.Str] = struct{}{}
+			}
+		}
+	}
+
+	// Handle initial cursor position if specified
+	if o.Initial != "" {
+		for i, match := range matches {
+			if match.Str == o.Initial {
+				m.cursor = i
+				break
 			}
 		}
 	}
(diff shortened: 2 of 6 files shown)


filter/command.go
```diff
@@ -132,15 +132,26 @@ func (o Options) Run() error {
 				continue
 			}
 			if o.Limit == 1 {
+				// When the user can choose only one option don't select the option but
+				// start with the cursor hovering over it.
```
sadamczyk, line 136:
```suggestion
			if o.Initial != "" && option.Str == o.Initial {
```

Currently `--selected` has two different behaviors.
1. If `--no-limit` or `--limit=<any value above 1>` then the cursor position isn't changed at all. This can stay.
2. If `--limit=1` (the default), then it only sets the initial cursor position. But that's exactly what the new option `--initial` does, so it does not make much sense to me to keep this behavior of `--selected` around. Because then there would be two ways to do the exact same thing, which seems confusing.
This also saves us from having to loop over all the options again in the second loop, so that can be removed.

This would be a significant BC break though, because I imagine a lot of people rely on this behavior of `--selected` because it's the default. So it might be better to deprecate this behavior of `--selected` first, before just outright removing it.

choose/options.go
```diff
@@ -21,6 +21,7 @@ type Options struct {
 	SelectedPrefix   string        `help:"Prefix to show on selected items (hidden if limit is 1)" default:"✓ " env:"GUM_CHOOSE_SELECTED_PREFIX"`
 	UnselectedPrefix string        `help:"Prefix to show on unselected items (hidden if limit is 1)" default:"• " env:"GUM_CHOOSE_UNSELECTED_PREFIX"`
 	Selected         []string      `help:"Options that should start as selected (selects all if given *)" default:"" env:"GUM_CHOOSE_SELECTED"`
+	Initial          string        `help:"Option that should start with the cursor positioned on it" default:"" env:"GUM_CHOOSE_INITIAL"`
```
sadamczyk, line 24: I personally feel like `--cursor` might be the better name for this new option.
Especially in combination with `--selected` it communicates very clearly that it only affects the cursor position, whereas `--inital` feels a little ambigous to me.
But as always naming things is one of the hardest problems :), so I'd be curious to hear other people's input on this.

sadamczyk (commented):
Thanks for taking a try at this!
In general it seemed to be doing exactly what I wanted, which is great! :)

PR #942, opened 2025-08-05
feat(spin): add support for custom spinner

### LINKED ISSUE #234: gum spin to support Fira Code spinner codepoints.
There are various spinning box ways in the terminal. Fira Code tries to spice that up a bit by providing actual spinning glyphs.

I would like `gum spin` to make use of that and provide  support for it via `gum spin --spinner=fira` with `fira` denoting that the Fira Code introduce spinning codepoints will be used (see link below).

ASCII arts

**Additional context**
See https://github.com/tonsky/FiraCode/releases/tag/6 for the release page, description & demo

Rakioth:
Fixes #234

### Changes
- Add `--glyphs` and `--frame-interval` flags to `gum spin` to support custom spinners.


spin/options.go
```diff
@@ -20,4 +20,6 @@ type Options struct {
 	TitleStyle   style.Styles  `embed:"" prefix:"title." envprefix:"GUM_SPIN_TITLE_"`
 	Align        string        `help:"Alignment of spinner with regard to the title" short:"a" type:"align" enum:"left,right" default:"left" env:"GUM_SPIN_ALIGN"`
 	Timeout      time.Duration `help:"Timeout until spin command aborts" default:"0s" env:"GUM_SPIN_TIMEOUT"`
+	Glyphs       []string      `help:"Characters or symbols used to animate the spinner (overrides spinner type)" default:"" env:"GUM_SPIN_GLYPHS"`
+	Fps          time.Duration `help:"Animation speed when using custom glyphs (lower is faster)" default:"100ms" env:"GUM_SPIN_FPS"`
```
ccoVeille, line 24: There is something wrong here

FPS state for frame per second.

And here you provide a duration. It's not logic.

Using Go duration is file and safer and simpler than using frame per second,that would imply to use a 1/x division.

So here I would suggest finding a better name for this parameter

Rakioth: You're right using Fps is a bit misleading maybe something like Delay?

```
Delay time.Duration `help:"Delay between animation frames when using custom glyphs (lower is faster)" default:"100ms" env:"GUM_SPIN_DELAY"`
```

ccoVeille: AnimationInterval maybe?

Rakioth: I like `interval`, but maybe `--animation-interval` is a bit long? What about `--frame-interval` instead?

spin/command.go
```diff
@@ -19,7 +19,16 @@ func (o Options) Run() error {
...
-	s.Spinner = spinnerMap[o.Spinner]
+
+	if len(o.Glyphs) > 0 {
+		s.Spinner = spinner.Spinner{
+			Frames: o.Glyphs,
+			FPS:    o.FrameInterval,
+		}
+	} else {
+		s.Spinner = spinnerMap[o.Spinner]
+	}
```
ccoVeille, line 30: I tend to avoid else when possible

```suggestion
	s.Spinner = spinnerMap[o.Spinner]
	if len(o.Glyphs) > 0 {
		s.Spinner = spinner.Spinner{
			Frames: o.Glyphs,
			FPS:    o.FrameInterval,
		}
	}
```

Issue #891, opened 2025-04-15
gum format adds many spaces to end of string


octoshrimpy:
**Describe the bug**
gum format -t emoji pads with many spaces

**To Reproduce**

```sh
echo 'I :heart: Bubble Gum :candy:' | gum format -t emoji
```
> I ❤️ Bubble Gum 🍬

```sh
echo 'I :heart: Bubble Gum :candy:' | gum format -t emoji | sed 's/ /-/g' 
```
> I-❤️-Bubble-Gum-🍬--------------------------------------------------------------

```sh
echo 'hi' | gum format -t emoji
```
> hi

```sh
echo 'hi' | gum format -t emoji | sed 's/ /-/g'
```
> hi------------------------------------------------------------------------------

**Expected behavior**
no extra spaces added

**Screenshots**

![Image](https://github.com/user-attachments/assets/d46ebe40-dc25-4b40-b2cd-8900832ce8ab)

> gum --version
> gum version 0.16.0 (0d116b8)

---

sidenote: looks like emoji aren't getting extra padding, as seen on screenie by the ❤ being so close to the `B`

PR #950, opened 2025-08-20
fix(format): remove trailing padding for emoji renderer

security-log:
Fixes https://github.com/charmbracelet/gum/issues/891

Changes
- Modified: format/formats.go
  - emoji(): include `glamour.WithWordWrap(0)`.

Why
- The renderer introduced unwanted padding when converting short lines containing emoji. Disabling word wrap prevents Glamour from adding that padding.

How to reproduce

Before the fix
```bash
echo 'I :heart: Bubble Gum :candy:' | ./gum format -t emoji | sed 's/ /-/g'
I-❤-Bubble-Gum-🍬--------------------------------------------------------------
```
After the fix
```bash
echo 'I :heart: Bubble Gum :candy:' | ./gum format -t emoji | sed 's/ /-/g'
I-❤-Bubble-Gum-🍬
```

### DIFF
--- a/format/formats_test.go
+++ b/format/formats_test.go
@@ -1,0 +1,69 @@
+package format
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestEmojiFormat_NoTrailingPadding(t *testing.T) {
+	input := "I :heart: Bubble Gum :candy:"
+	output, err := emoji(input)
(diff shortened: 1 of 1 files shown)


tauraamui (changes requested):
Hi! Can you add a test for this?

PR #953, opened 2025-08-22
feat: Adding --show-elapsed flag in spin command

### LINKED ISSUE #951: `gum spin` displays expired time

In some cases, e.g. long-running tests, it'd be useful to know how long a task has already been running.

An option of `gum --spin` that switches on display of expired time:

```
⣽  Running ... [3s]
⣽  Running ... [4s]
#...
⣽  Running ... [1:15m]
⣽  Running ... [1:16m]
```

None.

NucleoFusion:
Fixes #951 

### Changes
- adds a --show-elapsed flag that shows elapsed time.

image
image

``` go
	return fmt.Sprintf("(%dd %dh %dm)", days, hours, minutes)
	
	return fmt.Sprintf("(%dh %dm)", hours, minutes)
	
	return fmt.Sprintf("(%dm %ds)", minutes, seconds)

	return fmt.Sprintf("(%ds)", seconds)
```
These are the formats

### DIFF
--- a/spin/spin.go
+++ b/spin/spin.go
@@ -17,11 +17,13 @@ package spin
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"runtime"
 	"syscall"
+	"time"
 
 	"github.com/charmbracelet/bubbles/spinner"
 	tea "github.com/charmbracelet/bubbletea"
@@ -30,20 +32,22 @@ import (
 )
 
 type model struct {
-	spinner    spinner.Model
-	title      string
-	align      string
-	command    []string
-	quitting   bool
-	isTTY      bool
-	status     int
-	stdout     string
-	stderr     string
-	output     string
-	showStdout bool
-	showStderr bool
-	showError  bool
-	err        error
+	spinner     spinner.Model
+	title       string
+	align       string
+	command     []string
+	quitting    bool
+	isTTY       bool
+	status      int
+	stdout      string
+	stderr      string
+	output      string
+	showStdout  bool
+	showStderr  bool
+	showError   bool
+	showElapsed bool
+	err         error
+	started     time.Time
 }
 
 var (
@@ -55,6 +59,10 @@ var (
 )
 
 type errorMsg error
+
+type startMsg struct {
+	started time.Time
+}
 
 type finishCommandMsg struct {
 	stdout string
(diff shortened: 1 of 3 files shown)


spin/spin.go
```diff
@@ -137,10 +145,31 @@ func commandAbort() tea.Msg {
...
+	seconds := int(d.Seconds()) % 60
+
+	if days > 0 {
+		return fmt.Sprintf("(%dd %dh %dm)", days, hours, minutes)
+	} else if hours > 0 {
+		return fmt.Sprintf("(%dh %dm)", hours, minutes)
+	} else if minutes > 0 {
+		return fmt.Sprintf("(%dm %ds)", minutes, seconds)
+	}
+	return fmt.Sprintf("(%ds)", seconds)
```
reitzig, line 161: 💡 Maybe don't print anything if `seconds == 0`.

spin/spin.go
```diff
@@ -161,17 +190,25 @@ func (m model) View() string {
...
+	}
+
 	var header string
 	if m.align == "left" {
 		header = m.spinner.View() + " " + m.title
 	} else {
 		header = m.title + " " + m.spinner.View()
 	}
-	return header + "\n" + out
+	return header + elapsed + "\n" + out
```
reitzig, line 204: 💡 I think I'd appreciate a small separator between `header` and `elapsed`.  Just a single space, if there's nothing "better" in the framework.

With suffix `...` as in your example, it looks fine, but with text it may not.

spin/spin.go
```diff
@@ -137,10 +145,31 @@ func commandAbort() tea.Msg {
...
+
+	if days > 0 {
+		return fmt.Sprintf("(%dd %dh %dm)", days, hours, minutes)
+	} else if hours > 0 {
+		return fmt.Sprintf("(%dh %dm)", hours, minutes)
+	} else if minutes > 0 {
+		return fmt.Sprintf("(%dm %ds)", minutes, seconds)
+	}
+	return fmt.Sprintf("(%ds)", seconds)
+}
```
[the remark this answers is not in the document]
reitzig: No, I mean formatting the duration. I realize we should round, but this:


```go
var start = time.Now()
for true {
  time.Sleep(100 * time.Millisecond)
  fmt.Printf("%s\n", time.Since(start).Round(100*time.Millisecond))
}
```

creates output like

```
59.6s
59.7s
59.9s
1m0s
1m0.1s
1m0.2s
1m0.3s
```

Which is basically what you propose, sans the spaces. Hence my question, if you think the visual difference justifies the additional code.

(Not my call, just a thought.)

PR #1062, opened 2026-04-09
 filter: guard zero-match toggle selection

vonbai:
Changes

  - guard filter.ToggleSelection() when the filtered match set
    is empty or the cursor is out of range
  - add focused regressions for zero-match selection flows
  - keep the change scoped to filter/filter.go and filter/
    filter_test.go

  Why

  When filtering reduces the result set to zero, selection-
  toggle paths can still reach ToggleSelection() and index the
  current match unconditionally. This keeps zero-match toggle
  interactions safe and aligns them with the existing no-op
  behavior used by adjacent cursor movement paths.

  Verification

  - go test ./filter

  Coverage added

  - direct guard regression for zero-match ToggleSelection()
  - interactive zero-match toggle regression
  - interactive zero-match toggle-and-next regression

### DIFF
--- a/filter/filter.go
+++ b/filter/filter.go
@@ -444,6 +444,10 @@ func (m *model) CursorDown() {
 }
 
 func (m *model) ToggleSelection() {
+	if len(m.matches) == 0 || m.cursor < 0 || m.cursor >= len(m.matches) {
+		return
+	}
+
 	if _, ok := m.selected[m.matches[m.cursor].Str]; ok {
 		delete(m.selected, m.matches[m.cursor].Str)
 		m.numSelected--
--- a/filter/filter_test.go
+++ b/filter/filter_test.go
@@ -4,6 +4,9 @@ import (
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


kvnloo:
Live check on main 4d089f95: ToggleSelection still indexes m.matches[m.cursor] with no empty-set guard (filter/filter.go 460-468). CursorUp/CursorDown already return on len==0. #1105 is that panic.

This PR is the one I'd merge of the three open attempts (#1062 / #1128 / #1108): the guard matches those cursor helpers, plus a direct TestToggleSelectionWithNoMatches.

One nit on the Update tests: default Toggle is ctrl+@ (and starts disabled). tea.KeyMsg{Type: tea.KeyNull} after SetEnabled(true) does not match that binding, so TestUpdateIgnoresToggleWhenNoMatches never reaches ToggleSelection. TestUpdateIgnoresToggleAndNextWhenNoMatches does send tab, which is ToggleAndNext — that one actually exercises the panic path.

#1128 is the same three-line guard without tests. #1108 adds coverage but leaves the panic in.

I am not opening another PR.

PR #1065, opened 2026-04-14
fix: table --print header styles not applied correctly (#936)

### LINKED ISSUE #936: Gum Tables behave inccorectly
### Setup

Please complete the following information along with version numbers, if applicable.
 - OS Ubuntu 20.04
 - Shell Bash (/bin/bash)
 - Terminal Emulator MobaxTerm (lastest)
 - Terminal Multiplexer None Utilized

*Note: you might encounter rendering issues if your locale does not use
`UTF-8` encoding. Please check your locale (`locale` on POSIX systems) to
see what encoding is being used by your system.*

### To Reproduce

1) Create a bash array, in my case the array is initialized in the top level and declare -g.
2) Push to the array via for loop using += 
3) Utilize separator in my case I am using |
4) Specify --header.foreground="###" 
### Source Code

Please include a [minimum reproducible example][writing-example] to reproduce the behavior.
```bash
#!/bin/bash

# This was created as a means to reporting a bug to charm-bracelet/gum
# This will not include any sensitive data from the org.
# To be deleted after bug report resolved.

# First create/initialize array
declare -g gum_array
gum_array=()
# There are a couple of different ways we define the header, either by making it the first entry or by explicit --columns flag
function generate_gum_table() {
    method="$1"
    case $method in
        "first")
            printf "%s\n" "${gum_array[@]}" | gum table --separator="|" --print --columns="Account,User ID,Home Directory,Shell,Status" --border.foreground="240" --border="rounded" --header.background="240"
        ;;
        "second")
            printf "%s\n" "${gum_array[@]}" | printf "%s\n" "${gum_array[@]}" | gum table --separator="|" --print --border.foreground="240" --border="rounded" --header.background="240"
        ;;
        *)
            printf "%s\n" "${gum_array[@]}" | printf "%s\n" "${gum_array[@]}" | gum table --separator="|" --print --border.foreground="240" --border="rounded" 
        ;;
    esac
}
#  
#
# Not sure what to use for sample data lets just make a passwd parser
thefile=$(cat /etc/passwd)
while IFS= read -r line; do
    # Eval is the user currently locked? (as by their shell set to nologin)
    if [[ $(echo $line | grep "/usr/sbin/nologin") ]]; then
        gum_array+=( "$(echo "$line"| awk -F ':' '{print $1"|"$4"|"$6"|"$7"|Locked"}' | gum style --foreground="009")" ) 
    else
        gum_array+=( "$(echo "$line"| awk -F ':' '{print $1"|"$4"|"$6"|"$7"|Active"}' | gum style --foreground="002")" ) 
    fi
done <<< $thefile

# Now we printout our arrays value and pipe it in as standard input to the gum table command structure.
# Bug #1 Header is set as the first row of the table, whereby traditionally it is (atleast in 99% of programs) what defines the columns..
gum style --foreground="006" "Here is the first table, as you can see I have set the columns explicitly, both in color and in location but they do not go where expected."
generate_gum_table "first" # The first method is where we define the columns for the table explicitly
# Set headers via stin, by shifting the tables definition +1 whilst also injecting thje header.
gum_array=(
    "Account|User ID|Home Directory|Shell|Status"
    "${gum_array[@]}"
)
gum style --foreground="006" "Here is the second table, as you can see I have relied on stin, both in color and in location but they do not go where expected."
generate_gum_table "second" # here we're going to rely on first entry.
# Bug #2, this is in bug #1 but we want to make sure it is adequately identified. The first row despite the clearly defined out color as 002 or 009 will show brighter with no header flags at all.
gum style --foreground="006" "Here is the third table, what I am trying show here is the color for the first row in the table is lighter, but I have not set the color to be brighter."
generate_gum_table # There will be a lighter shade of a color even through I defined it before stin. 
gum style --foreground="006" "Thank you for your time, and looking forward to what you cook up!"
```

Please paste your example as a code block on the issue. Do not link to an external project. Your example should fit in a single file. Take a look at [Bubble Tea examples][bubbletea-example] for an idea on writing a small, but runnable program.

[bubbletea-example]: https://github.com/charmbracelet/bubbletea/tree/main/examples
[writing-example]: https://stackoverflow.com/help/minimal-reproducible-example

### Expected behavior

The headers are never correctly setup, for some reason the command is using row 1 as its header row instead of the actual header row.

The color for the first row is adjusted even through it was defined before hand.

### Screenshots

Image

Image

Add screenshots to help explain your problem.

### Additional context

Looking forward to your response!

HoCoOhNo:
Fixes #936

### Changes
- header style applied to row -1 instead of 0

gum table --print was applying header style to the first row of data, after the headers, not the headers. Changed this value to -1 and it works as intended

### DIFF
--- a/table/command.go
+++ b/table/command.go
@@ -117,7 +117,7 @@ func (o Options) Run() error {
 			BorderStyle(o.BorderStyle.ToLipgloss()).
 			Border(style.Border[o.Border]).
 			StyleFunc(func(row, _ int) lipgloss.Style {
-				if row == 0 {
+				if row == -1 {
 					return styles.Header
 				}
 				return styles.Cell


HoCoOhNo:
Does this look ok? Can it be merged?

PR #1089, opened 2026-06-18
fix(style): support 3-value padding/margin shorthand

s3onghyun:
### Changes

- Fix `--padding`/`--margin` with **3 values** being silently ignored. `ParsePadding` (`style/spacing.go`) handled 1, 2, and 4 tokens but had no branch for 3, so `gum style --padding "1 2 3" "text"` fell through to `(0,0,0,0)` and rendered with no padding.
- Add the missing 3-token branch using the CSS shorthand lipgloss already supports — top / horizontal / bottom → `(top, horizontal, bottom, horizontal)`.
- Add `style/spacing_test.go` with a table test for `ParsePadding` (1/2/3/4 tokens). The 3-token case fails before this change and passes after; `go test ./...` green.

### DIFF
--- a/style/spacing.go
+++ b/style/spacing.go
@@ -6,9 +6,10 @@ import (
 )
 
 const (
-	minTokens  = 1
-	halfTokens = 2
-	maxTokens  = 4
+	minTokens   = 1
+	halfTokens  = 2
+	threeTokens = 3
+	maxTokens   = 4
 )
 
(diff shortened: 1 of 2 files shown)


s3onghyun:
Closing this to refocus my open-source time on a smaller set of projects. The change still stands on its own — happy to reopen if a maintainer would like to pick it up. Thanks for maintaining this!

PR #1112, opened 2026-08-06
chore: update to bubbletea v2

andrinoff:
Updates Charm libraries to v2


go.mod
andreynering, line 1: Reminder that we'll probably want to bump this to v2 when releasing!


```suggestion
module charm.land/gum/v2
```

andreynering (approved):
👏

go.mod
andrinoff: will do!

PR #1129, opened 2026-08-25
Fix handling of the Space-key in `choose` dialogs

alxndrv:
Couldn't find an existing Github issue for this.

We recently got reports on some internal scripts saying that pressing Space was no longer toggling options in a `gum choose` invocation (Tab and X working just fine).

This seems to be some form of regression that came with the v2 release:
- `gum choose --no-limit "one" "two" "three"` on the latest version, pressing Space does nothing.
- Running the same command on v0.17.0, pressing Space toggles options as expected.

### Changes
- Updated the keybinding definition from `" "` to `"space"`, which seemingly fixes the aforementioned issue 🤷


ebryan-path:
One addition I would suggest while you're here: the `ctrl+@` keybind is supposed to mean `ctrl+space`, and that's something else which used to work but broke in v2. I believe the fix is simple, just replace occurrences of `ctrl+@` with `ctrl+space` here and in filter/filter.go - those seem to be the only places which use this binding.

alxndrv:
> One addition I would suggest while you're here: the `ctrl+@` keybind is supposed to mean `ctrl+space`, and that's something else which used to work but broke in v2. I believe the fix is simple, just replace occurrences of `ctrl+@` with `ctrl+space` here and in filter/filter.go - those seem to be the only places which use this binding.

Thanks for the pointer, didn't realize that's what those `ctrl+@` instances were referring to. 🙏 
Pushed a change with those included (seems to work fine now on my machine).

ebryan-path:
Happy to help! When I noticed that the space binding had broken and found your PR here, I checked the other bindings to see if they had broken too. I had to dig to understand what `ctrl+@` meant - as I understand it, both `ctrl+space` and `ctrl+@` (and sometimes `ctrl+2`) usually get mapped to a NUL character by terminals, so the old `ctrl+@` is apparently how gum used to express "trigger when you receive a NUL character", and presumably the actual intent there was to let both `space` and `ctrl+space` toggle items.

PR #1134, opened 2026-08-28
fix(format): preserve colors when piped to fix v2 regression

### LINKED ISSUE #1131: `gum format` loses colors when  piped (new in v2)
### Description

in v1 one could do `gum format | less -R` and have both color formatting and pagination
v2 broke this

### Version

_No response_

### Environment

_No response_

marwan562:
Fixes #1131

### Changes
- In `format/command.go:46-67` preserve ANSI colors when stdout is piped (e.g. `gum format | less -R`, `| cat`)
- Replace `tty.Writer()` (`colorprofile.Detect` → NoTTY strips when not TTY) with `colorprofile.Env` (ignores TTY, respects `NO_COLOR`, `CLICOLOR_FORCE`, `TERM`/`COLORTERM`)
- Fallback `NoTTY` → `ANSI` when `TERM` missing (v1 `format | less -R` still colored), respect `TERM=dumb` → still strip unless `CLICOLOR_FORCE`
- Uses `colorprofile.Writer{Forward: os.Stdout, Profile: profile}` to downgrade (TrueColor→256→ANSI) appropriately instead of stripping

### Root cause
v2 `format/command.go:45` used `internal/tty.Writer()` → `colorprofile.NewWriter(os.Stdout)` → `Detect` checks `term.IsTerminal`. When piped `isatty=false` → profile `NoTTY` → `Writer.Write` does `ansi.Strip`, removing all colors. v1 wrote directly (no profile), so `format | less -R` kept colors. Workaround `CLICOLOR_FORCE=1` forced `ANSI` in `Detect`, restoring colors.

### Validation
- [x] `go vet ./format` / `golangci-lint run ./format` 0 issues
- [x] `go test ./...` pass
- [x] Reproduced before: `echo "# Hello" | gum format --theme pink | cat | od -c` had no ESC; with `CLICOLOR_FORCE=1` had ESC 38;5;212
- [x] After fix: `echo "# Hello" | gum format | cat | od -c` has ESC 38;5;212 (piped preserves), `NO_COLOR=1 …| cat` strips colors (keeps bold ESC 1m), `TERM=dumb …| cat` strips, `echo 'func main(){}' | gum format --type code | cat` has ESC


akbloch-dn:
thanks for considering this issue.

other than a clever identification of whether the tty handles colors, and other than env var enforcing - I think a cli flag such as `--colors=<always/never/auto>` also makes some sense
to your judgement

marwan562:
Hey @akbloch-dn, thanks for the suggestion!

You're right — `--color=always/never/auto` is the classic pattern (like `ls`, `bat`) and more discoverable than env vars, so makes sense.

I kept the fix minimal and just pushed an update:

- `auto` (default) — keeps the regression fix: preserves colors when piped (`gum format | less -R`) via `colorprofile.Env`, still respects `NO_COLOR`, `CLICOLOR_FORCE` and `TERM`. So `format | cat` works again like in v1.
- `always` — forces at least ANSI even with `TERM=dumb` or `NO_COLOR`
- `never` — fully strips ANSI (a bit stronger than `NO_COLOR` which keeps bold)

Env vars still work the same (`CLICOLOR_FORCE=1` for always, `NO_COLOR=1` for never), the flag is just an explicit, per-invocation alternative.

I went with `--color` (singular) to match the usual naming, but happy to rename to `--colors` if you prefer, and `--color` default stays `auto` so existing behavior doesn't break.

Let me know what you think — happy to tweak the naming/default if maintainers have a preference.

Thanks again for catching the v2 regression!

erik434:
Just my 2 cents as a fellow gum user: I suspect the best fix for this would be in the underlying colorprofile library, since it's not just `gum format` which lost color piping in v2, it's every gum command.

Or if this was an intentional change in colorprofile but not for gum, perhaps a tweak in `internal/tty/tty.go` where I assume all gum commands set up `stdout` using `colorprofile.NewWriter`. A fix there would presumably apply to all gum commands.

PR #1142, opened 2026-09-03
fix: force TrueColor profile when CLICOLOR_FORCE=1 (issue #1127)

### LINKED ISSUE #1127: Color output in v2 now requires a tty?
### Description

After updating to the latest gum v2 (installed via Homebrew), I noticed a change in behavior and I'd like to clarify if this was intentional or not. This is in bash 5.3.15 if it matters for the below examples. First, these commands work fine in both the old and new versions of gum:
```sh
gum style --foreground "#F00" hello
(gum style --foreground "#F00" hello)
```

They print "hello" in red text as expected. The following line has the same behavior in the previous version (gum v0.17.0), but since upgrading to v2, it now prints "hello" without color:
```sh
echo "$(gum style --foreground "#F00" hello)"
```

It appears that v2 doesn't print colors unless it detects a tty(?), but the older version didn't care. Relevant environment variables I have set:
```
COLORTERM=truecolor
TERM=xterm-256color
```

Was this intentional or is this a bug? If it's intentional, is there a way to tell `gum` to output color anyway? I didn't find anything in the documentation about enabling/disabling color output.

### Version

v2.0.0

### Environment

Ubuntu 20.04.6 / mobaxterm v25.0

dajiaohuang:
## Description

Fix for issue #1127: Color output in gum v2 requires a TTY.

In gum v1, color output worked in subshells and command substitution. In v2, the color output requires a TTY due to the colorprofile package detecting NoTTY profile when stdout is not a terminal.

## Root Cause

When running `echo "$(gum style --foreground "#F00" hello)"` in a subshell, stdout is not a TTY. The colorprofile package detects this and returns a NoTTY profile, which strips all ANSI sequences.

While the colorprofile package respects CLICOLOR_FORCE=1, it only upgrades the profile to ANSI (not TrueColor), which still causes TrueColor sequences to be downsampled.

## Fix

Modified `internal/tty/tty.go` to force TrueColor profile when CLICOLOR_FORCE=1 is set. This ensures that ANSI sequences are preserved (not stripped or downsampled) when the user explicitly requests color output via CLICOLOR_FORCE.

## Changes

- `internal/tty/tty.go`: Added check for CLICOLOR_FORCE=1 to force TrueColor profile
- `internal/tty/tty_test.go`: Added regression test for the fix

## Testing

The fix can be verified by running:

```bash
CLICOLOR_FORCE=1 echo "$(gum style --foreground "#F00" hello)"
```

This should now output red colored text (with ANSI escape sequences) instead of plain text.

---

Fixes #1127


erik434:
I'm not sure this is needed. From what I can see, `CLICOLOR_FORCE=1` enables color when it would normally be disabled, but it doesn't restrict you to ANSI - that's just the default. If you also set `COLORTERM=truecolor`, you should still get truecolor output. Examples:

```bash
export CLICOLOR_FORCE=1
export COLORTERM=truecolor

echo "$(gum style --foreground "#FF0000" hello)" | cat -v
# prints: ^[[38;2;255;0;0mhello^[[m
echo "$(gum style --foreground "#FF0102" hello)" | cat -v
# prints: ^[[38;2;255;1;2mhello^[[m

export COLORTERM=

echo "$(gum style --foreground "#FF0102" hello)" | cat -v
# prints: ^[[38;5;196mhello^[[m
```
As you can see, the first two outputs include `38;2`, which seems to indicate truecolor, followed by the RGB values. You can see in the first case they are 255/0/0 as you would expect for FF0000, and in the second case they are also as expected, 255/1/2 for FF0102.

The last case, with COLORTERM cleared, prints `38;5` for ANSI 256 color mode, then color index 196, which is a pinkish red, probably the closest color in the palette to the requested FF0102.

I believe you're probably seeing this last case - an environment without COLORTERM defaults to ANSI, so you probably just need to set `COLORTERM=truecolor`.

Side note, in your test case `CLICOLOR_FORCE=1 echo "$(gum style --foreground "#F00" hello)"`, the CLICOLOR_FORCE only applies to `echo`, so it doesn't have any effect on `gum`. If you run `echo "$(CLICOLOR_FORCE=1 gum style --foreground "#F00" hello)"` you'll see red text even with the current v2 of gum.

PR #1152, opened 2026-09-14
fix(filter): prevent panic in ToggleSelection with empty matches

### LINKED ISSUE #1105: `gum filter` panics (index out of range) when toggling selection with no matches
### Description

When using `gum filter` in multi-select mode (`--limit` > 1 or `--no-limit`), typing a search query that matches zero options and then pressing Tab (or Ctrl+@ / Shift+Tab) crashes the program with an `index out of range` panic instead of doing nothing.

**To Reproduce**
1. Run `printf 'apple\nbanana\ncherry\n' | gum filter --no-limit`
2. Type a query that matches nothing, e.g. `zzz`
3. Press `Tab`

**Expected behavior**
Pressing Tab with no visible matches should be a no-op (same as `CursorUp`/`CursorDown` already handle this case), not crash the program.

**Actual behavior**
```
Caught panic: runtime error: index out of range [0] with length 0
```
(or similar)

**Additional context**
Root cause: `ToggleSelection()` in `filter.go` indexes into `m.matches[m.cursor]` without checking `len(m.matches) == 0` first:

```go
func (m *model) ToggleSelection() {
	if _, ok := m.selected[m.matches[m.cursor].Str]; ok {   // panics when m.matches is empty
	...
```

By contrast, `CursorUp()` and `CursorDown()` both guard against this:
```go
func (m *model) CursorUp() {
	if len(m.matches) == 0 {
		return
	}
	...
```

**Suggested fix**
Add the same empty-matches guard to `ToggleSelection`:
```go
func (m *model) ToggleSelection() {
	if len(m.matches) == 0 {
		return
	}
	...
```

### Version

v0.17.0

### Environment

Ghostty

rksharma-owg:
Fixes #1105

### Problem
When using `gum filter` in multi-select mode (`--limit > 1` or `--no-limit`), typing a search query that matches zero options and then pressing Tab (or toggle key) causes a runtime panic: `index out of range [0] with length 0`.

Root cause: `ToggleSelection()` in `filter/filter.go` indexed directly into `m.matches[m.cursor]` without checking whether `m.matches` was empty or `m.cursor` was valid.

### Fix
- Add a guard check to `ToggleSelection()`: `if len(m.matches) == 0 || m.cursor < 0 || m.cursor >= len(m.matches) { return }`, consistent with the bounds check in `CursorUp()` and `CursorDown()`.
- Add unit tests in `filter/filter_test.go` covering `ToggleSelection` when `matches` is empty, when the cursor is out of bounds, and during normal toggle/untoggle.

### Verification
- `go test -v ./filter` passes cleanly
- `go test ./...` passes cleanly


otrumb (approved):
Verified at 70c936d4bf68f3fe66525e162b670f145bc76062. The bounds guard fixes #1105: zero-match toggles now leave selection unchanged, while filtering and repopulation preserve usable cursor and selection behavior. A regression probe panics on the base commit and passes on this head across all three toggle bindings, fuzzy and exact matching, and both layouts. Native Windows focused/full tests, go vet, and go build passed with CGO disabled. No blocking findings.

